PR #7, opened 2015-04-09
Get rid of Volley

daniloercoli:
This PR gets rid of Volley network lib, and switches back to the native HttpURLConnection.

Getting rid of Volley we should see low memory consumption, and better performances. We don't need to use Volley to make a simple POST request to the server. Volley adds nice features to the networking layer, but a bit of overkill for this project. Memory consumption and overall performance are better with HttpURLConnection for the kind of network connection we have in this prj (single connection, no pooling is require, no tagging of connection required, etc).


daniloercoli:
//cc @astralbodies @maxme

astralbodies:
> Showing  4 changed files  with 170 additions and 269 deletions.
Music to my ears.

Functionally this is in line with what I did with the iOS client - tried to use base libraries/frameworks as much as possible.

maxme:
` mQueue = Volley.newRequestQueue(ctx);` can be removed

maxme:
Volley dependency should be removed in the `build.gradle` file

```
dependencies {
    compile 'com.mcxiaoke.volley:library:1.0.15'
}
```

maxme:
PR looks fine, we just need to remove the gradle dependency before the merge.

daniloercoli:
Removed the gradle dependency and got rid of unused variables.

PR #9, opened 2015-04-17
Use the correct regExp to validate properties

### LINKED ISSUE #8: Use the proper regexp to validate properties
We should use a slightly different regexp to validate properties. At the moment we are using one regExp for both events and properties.

daniloercoli:
Fix #8 

//cc @astralbodies


astralbodies:
Looks good to me :smile: 

Making a similar change in the iOS client and the unit tests passed with this Regex.

astralbodies:
:shipit:

PR #15, opened 2015-06-23
Do not immediataly retry on network errors

### LINKED ISSUE #14: Do not immediataly retry on network error
We should add a back-off time on network request error, otherwise there is a very high possibility to fall in a loop where the network connection is retried every second.

daniloercoli:
Fix #14 by adding a fixed time of 2 minutes before retrying the network request. Otherwise there is a very high possibility to fall in a loop where the network connection is retried every second.

Also bumps version number to 1.0.4

### DIFF
--- a/AutomatticTracks/build.gradle
+++ b/AutomatticTracks/build.gradle
@@ -25,7 +25,7 @@ android {
     buildToolsVersion "21.1.2"
 
     defaultConfig {
-        versionName "1.0.3"
+        versionName "1.0.4"
         minSdkVersion 14
         targetSdkVersion 19
     }
--- a/AutomatticTracks/build.gradle
+++ b/AutomatticTracks/build.gradle
@@ -21,13 +21,13 @@ dependencies {
 android {
     publishNonDefault true
 
-    compileSdkVersion 19
-    buildToolsVersion "21.1.2"
+    compileSdkVersion 22
+    buildToolsVersion "22.0.1"
 
     defaultConfig {
         versionName "1.0.4"
         minSdkVersion 14
-        targetSdkVersion 19
+        targetSdkVersion 22
     }
 
     compileOptions {
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/TracksClient.java
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/TracksClient.java
@@ -63,7 +63,7 @@ public class TracksClient {
 
     private boolean mPendingFlush = false;
     private static long WAIT_PERIOD_NETWORK_CONNECTION = 2 * 60 * 1000 ; // 2 Minutes
-    private long lastNetworkErrorTimestamp = 0L;
+    private long mLastNetworkErrorTimestamp = 0L;
 
     public static TracksClient getClient(Context ctx) {
         if (null == ctx || !checkBasicConfiguration(ctx)) {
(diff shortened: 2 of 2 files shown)


AutomatticTracks/src/main/java/com/automattic/android/tracks/TracksClient.java
```diff
@@ -267,6 +274,8 @@ public void run() {
                                         mInsertEventsQueue.notifyAll();
                                     }
                                 }
+                            } else {
+                                lastNetworkErrorTimestamp = 0L;
```
tonyr59h: Could it be the case that this value is overwritten before it's being compared?

For example:
1. We set the error timestamp to currentTimeMs (262)
2. The bufferCopyThread begins inserting the mustKeepEvents to the DB (88)
3. The network thread processes another event with success, sets lastNetworkErrorTimestamp to 0 (278)
4. sendToWireThread begins processing events in the database (121)
5. lastNetworkErrorTimestamp is 0 -> shouldWait is false (125)
6. There is no pending flush so the events are not enqueued in the network thread (127)

Would be bizarre to have the network thread process another message before the sendToWire thread starts running but it looks like it's possible.

I think this will prevent that with another side-effect that the retry will happen two minutes from the first error instead of the last error.
```
if ((mPendingFlush || (!shouldWait && EventTable.getEventsCount(mContext) > DEFAULT_EVENTS_QUEUE_THREESHOLD))
                                    && NetworkUtils.isNetworkAvailable(mContext)) {
    lastNetworkErrorTimestamp = 0L;
    ...
}

if (lastNetworkErrorTimestamp == 0L && isErrorResponse) {
    lastNetworkErrorTimestamp = System.currentTimeMillis();
    ...
}
```

AutomatticTracks/src/main/java/com/automattic/android/tracks/TracksClient.java
```diff
@@ -62,7 +62,8 @@
 
 
     private boolean mPendingFlush = false;
-
+    private static long WAIT_PERIOD_NETWORK_CONNECTION = 2 * 60 * 1000 ; // 2 Minutes
+    private long lastNetworkErrorTimestamp = 0L;
```
tonyr59h: Rename to mLastNetworkErrorTimestamp for convention.

AutomatticTracks/build.gradle
```diff
@@ -25,7 +25,7 @@ android {
     buildToolsVersion "21.1.2"
 
     defaultConfig {
-        versionName "1.0.3"
+        versionName "1.0.4"
         minSdkVersion 14
         targetSdkVersion 19
```
tonyr59h: We should bump the target and compile SDK version to 22.

AutomatticTracks/src/main/java/com/automattic/android/tracks/TracksClient.java
daniloercoli: I'm happy that you've found a flow path that could lead to a weird behavior. 
Even if I'm not sure I've 100% understood the flow above :)
It's unlikely that the network thread process other messages immediately after an error, because it synchronizes over `mNetworkQueue` that should be empty at that point. 
`mNetworkQueue` usually contains 1 element only, with all the events that should be sent over the wire with a single REST call.

If I got it right, the problem you've mentioned could happen if the app sends 10+ events to the library, immediately after that `sendToWireThread` has written to the queue. In that case the `networkThread` will find another event in the queue and start to precess it without waiting the timeout.
I think it's a rare case, and never get into it during the tests.

tonyr59h: > If I got it right, the problem you've mentioned could happen if the app sends 10+ events to the library

That's exactly it.

> mNetworkQueue usually contains 1 element only, with all the events that should be sent over the wire with a single REST call.

Because of this I think we're safe. Even if it was the case that many elements are in the queue this code-path could only hit if the network latency was somehow faster than a simple thread-wake. So...it would likely never occur. Looks like it's good for a merge :+1:

tonyr59h:
:shipit:

tonyr59h:
I'll release an update today and get the WPAndroid PR merged in as well.

PR #17, opened 2015-07-22
Send events more often to the server

### LINKED ISSUE #16: Send events to the server more often
We've had our first step in comparing Mixpanel and Track numbers, and seems that Tracks numbers are always lower (5% of delta) than the ones recorded in Mixpanel.

The Android library works in this way:
- We store events in the db, when the fixed limit of 10 events is reached, all events are read from the DB and sent to the server with a batch request. If the request fails events are re-enqueued in the DB.
- When the app is sent in background, all events are sent to the server. Doesn't matter how many events are in the queue. 

My hypothesis is that the app fails to push events (ie: a network error) when the "on background" event  happens, and then the user doesn't reopen the app in the next 5 days.
<b>Important note:  On the server events that are 5 days old or more are rejected.</b>

How to fix this:
1) We can create a service that runs in background and sends the events to the server, but we will have hard time justifying a background process just to send usage data over the Internet.
2) Be more aggressive, and send events to the server more often. ie: Change the 10 events limit, and lower it to 5.

Also removing the policy that 5 days old events are discarded on the server may help a lot.

cc @martinremy @astralbodies @xyu

daniloercoli:
This PR should fix #16 by sending events to the server more often by using a timer, that is triggered also in case of network errors.

> My hypothesis is that the app fails to push events (ie: a network error) when the "on background" event happens, and then the user doesn't reopen the app in the next 5 days. 

This PR also bumps version number to 1.0.5. 

<b>Once this is merged we need to release the new version of the library on Maven</b>

### DIFF
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/TracksClient.java
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/TracksClient.java
@@ -29,7 +29,7 @@ public class TracksClient {
     public static final String LIB_VERSION = BuildConfig.VERSION_NAME;
     protected static final String DEFAULT_USER_AGENT = "Nosara Client for Android" + "/" + LIB_VERSION;
     protected static final String NOSARA_REST_API_ENDPOINT_URL_V1_1 = "https://public-api.wordpress.com/rest/v1.1/";
-    protected static final int DEFAULT_EVENTS_QUEUE_THREESHOLD = 10;
+    protected static final int DEFAULT_EVENTS_QUEUE_THREESHOLD = 5;
 
     public static enum NosaraUserType {ANON, WPCOM}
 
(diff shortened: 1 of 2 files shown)


AutomatticTracks/src/main/java/com/automattic/android/tracks/TracksClient.java
```diff
@@ -23,13 +24,16 @@
...
 
+
+
 public class TracksClient {
     public static final String LOGTAG = "NosaraClient";
 
     public static final String LIB_VERSION = BuildConfig.VERSION_NAME;
     protected static final String DEFAULT_USER_AGENT = "Nosara Client for Android" + "/" + LIB_VERSION;
     protected static final String NOSARA_REST_API_ENDPOINT_URL_V1_1 = "https://public-api.wordpress.com/rest/v1.1/";
-    protected static final int DEFAULT_EVENTS_QUEUE_THREESHOLD = 10;
```
tonyr59h: nitpick; should be DEFAULT_EVENTS_QUEUE_**THRESHOLD**

AutomatticTracks/src/main/java/com/automattic/android/tracks/TracksClient.java
```diff
@@ -23,13 +24,16 @@
...
+    protected static final int DEFAULT_EVENTS_QUEUE_THREESHOLD = 9;
```
tonyr59h: The bump to 9 seems arbitrary. Just testing the waters on lowering the limit?

daniloercoli: Hehe, it seems that I'm testing the water, but we want to send events in batch of 10 (and not 11), so the change.

tonyr59h:
:shipit:

PR #18, opened 2015-10-09
Add a User Type for Simplenote

roundhill:
Added `SIMPLENOTE` to the user type enum, and changed the message builder to always send the user type in lower case. 

I'm concerned I might have broken WPAndroid reporting, but it looks like it sets the user type to `WPCOM` anyways.

Idea for the future: maybe the library should allow clients to set the type string themselves?

@daniloercoli needs review!

### DIFF
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/MessageBuilder.java
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/MessageBuilder.java
@@ -105,8 +105,8 @@ class MessageBuilder {
                 eventJSON.put(USER_TYPE_KEY, USER_TYPE_ANON);
                 eventJSON.put(USER_ID_KEY, event.getUser());
             } else {
+                eventJSON.put(USER_TYPE_KEY, event.getUserType().name().toLowerCase());
                 eventJSON.put(USER_LOGIN_NAME_KEY, event.getUser());
-                // no need to put the user type key here. default wpcom is used on the server. 'wpcom:user_id'
             }
 
             unfolderPropertiesNotAvailableInCommon(event.getUserProperties(), USER_INFO_PREFIX, eventJSON, commonProps);
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/MessageBuilder.java
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/MessageBuilder.java
@@ -102,16 +102,18 @@ class MessageBuilder {
             }
 
             eventJSON.put(EVENT_TIMESTAMP_KEY, event.getTimeStamp());
-            eventJSON.put(USER_ID_KEY, event.getUser());
 
             switch (event.getUserType()) {
                 case ANON:
+                    eventJSON.put(USER_ID_KEY, event.getUser());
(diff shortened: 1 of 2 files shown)


AutomatticTracks/src/main/java/com/automattic/android/tracks/MessageBuilder.java
```diff
@@ -105,8 +105,8 @@ public static synchronized JSONObject createEventJSONObject(Event event, JSONObj
                 eventJSON.put(USER_TYPE_KEY, USER_TYPE_ANON);
                 eventJSON.put(USER_ID_KEY, event.getUser());
             } else {
+                eventJSON.put(USER_TYPE_KEY, event.getUserType().name().toLowerCase());
```
daniloercoli: I think that this code here `event.getUserType().name().toLowerCase()` may result in an invalid user type for WordPress.com users. In that case the string expected by the server is `wpcom:user:id`.
Should we change this code with a switch statements?

Did you add the usertype `simplenote` to the REST Endpoint?

roundhill: > wpcom:user:id, instead we're sending wpcom

Ah, see I thought I probably broke it ;) Thanks.

+1 for a switch statement! I'll change it to that

roundhill:
@daniloercoli ready for another look. I specifically sent the string `wpcom:user:id` for the WPCOM user type. Is that what you meant?

I'm also not certain if there is now a simplenote anon type... I saw some chat about that.

AutomatticTracks/src/main/java/com/automattic/android/tracks/MessageBuilder.java
```diff
@@ -18,7 +18,9 @@
     private static final String EVENT_TIMESTAMP_KEY = "_ts";
     private static final String REQUEST_TIMESTAMP_KEY = "_rt";
     private static final String USER_TYPE_KEY = "_ut";
-    private static final String USER_TYPE_ANON= "anon";
+    private static final String USER_TYPE_ANON = "anon";
+    private static final String USER_TYPE_WPCOM = "wpcom:user:id";
```
daniloercoli: This should be `wpcom:user_id`

AutomatticTracks/src/main/java/com/automattic/android/tracks/MessageBuilder.java
```diff
@@ -18,7 +18,9 @@
...
+    private static final String USER_TYPE_SIMPLENOTE = "simplenote";
```
daniloercoli: Same here, `simplenote:user_id`

daniloercoli:
> I'm also not certain if there is now a simplenote anon type... I saw some chat about that. 

No need to add another anon type.

AutomatticTracks/src/main/java/com/automattic/android/tracks/MessageBuilder.java
roundhill: Corrected in https://github.com/Automattic/Automattic-Tracks-Android/commit/743e22d0c7e3eca09a04c45a01316c7cdaa9ce39

daniloercoli:
:shipit:

PR #23, opened 2015-11-04
Fix crash while accessing BT on 4.x

### LINKED ISSUE #20: Crash report: RuntimeException in com.automattic.android.tracks.DeviceInformation
```
Fatal Exception: java.lang.RuntimeException: Can't create handler inside thread that has not called Looper.prepare()
       at android.os.Handler.(Handler.java)
       at android.bluetooth.BluetoothAdapter$2.(BluetoothAdapter.java)
       at android.bluetooth.BluetoothAdapter.(BluetoothAdapter.java)
       at android.bluetooth.BluetoothAdapter.getDefaultAdapter(BluetoothAdapter.java)
       at com.automattic.android.tracks.DeviceInformation.isBluetoothEnabled(DeviceInformation.java)
       at com.automattic.android.tracks.DeviceInformation.getMutableDeviceInfo(DeviceInformation.java)
       at com.automattic.android.tracks.MessageBuilder.createRequestCommonPropsJSONObject(MessageBuilder.java)
       at com.automattic.android.tracks.TracksClient$2.run(TracksClient.java)
       at java.lang.Thread.run(Thread.java)
```

daniloercoli:
Fix #20 and bump version number to 1.1

Once this is merged we need to publish it on Sonatype.

cc /@maxme
@roundhill  - This is the version you need to use in Simplenote.

### DIFF
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/DeviceInformation.java
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/DeviceInformation.java
@@ -242,6 +242,12 @@ import java.util.Locale;
     }
 
     public Boolean isBluetoothEnabled() {
+        // Do not retrieve bluetooth info on older Android versions.
+        // See: https://github.com/Automattic/Automattic-Tracks-Android/issues/20
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
+            return null;
+        }
+
         Boolean isBluetoothEnabled = null;
         if (PackageManager.PERMISSION_GRANTED == mContext.checkCallingOrSelfPermission(Manifest.permission.BLUETOOTH)) {
             BluetoothAdapter bluetoothAdapter = BluetoothAdapter.getDefaultAdapter();
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/DeviceInformation.java
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/DeviceInformation.java
@@ -243,13 +243,11 @@ import java.util.Locale;
 
     public Boolean isBluetoothEnabled() {
         Boolean isBluetoothEnabled = null;
-        try {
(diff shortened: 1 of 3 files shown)


AutomatticTracks/src/main/java/com/automattic/android/tracks/DeviceInformation.java
```diff
@@ -242,14 +242,18 @@ public Boolean isWifiConnected() {
...
     public Boolean isBluetoothEnabled() {
+        // Do not retrieve bluetooth info on older Android versions.
+        // See: https://github.com/Automattic/Automattic-Tracks-Android/issues/20
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
+            return null;
+        }
+
         Boolean isBluetoothEnabled = null;
-        try {
+        if (PackageManager.PERMISSION_GRANTED == mContext.checkCallingOrSelfPermission(Manifest.permission.BLUETOOTH)) {
```
maxme, line 252: From the doc:

> checkCallingOrSelfPermission: Determine whether the calling process of an IPC or you have been granted a particular permission. This is the same as checkCallingPermission(String), except it grants your own permissions if you are not currently processing an IPC. Use with care!

I think it's safer to keep the old code (catch `SecurityException`)

daniloercoli, line 252: We're already using `checkCallingOrSelfPermission` in other part of the lib without issues. It's not opening a security issue and offers the ability to access the library in IPC cases (that's not our case atm).

 I went beyond the doc since it's not so clear on this, and found out that in our cases (Tracks and WPAndroid) the call is identical to `checkPermission`. If we open the library/app to IPC we will enable the caller to access the BT info (if our process has the ability to access BT). We can change it to `checkPermission` safely though.

maxme, line 252: Ok, I don't have a strong opinion on this so :shipit:

PR #28, opened 2017-04-13
Fix null event names and event properties 

daniloercoli:
This PR fix an issue where passing an empty user to the library generates a bad crash.
I've also added other checks over the properties passed to the event, so the library should not crash if an invalid input is given.

- Also modernized gradle, and SDK.

Version number is also bumped, so we can release a new version of the library when this is merged to develop.

cc @maxme

### DIFF
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/datasets/EventTable.java
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/datasets/EventTable.java
@@ -102,7 +102,7 @@ public class EventTable {
             stmt.execute();
 
             db.setTransactionSuccessful();
-        } catch(IllegalArgumentException e) {
+        } catch (IllegalArgumentException e) {
             Log.e(TracksDatabaseHelper.LOGTAG, "Cannot insert the current event. Please check the details of the event!", e);
         } finally {
             db.endTransaction();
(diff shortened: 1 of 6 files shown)


AutomatticTracks/src/main/java/com/automattic/android/tracks/datasets/EventTable.java
```diff
@@ -97,11 +97,13 @@ public static void insertEvent(Context ctx, Event event) {
             }
 
             stmt.bindLong(8, event.getTimeStamp());
-            stmt.bindLong(9,  event.getRetryCount());
+            stmt.bindLong(9, event.getRetryCount());
 
             stmt.execute();
 
             db.setTransactionSuccessful();
+        } catch(IllegalArgumentException e) {
```
maxme: I think we should add our checkstyle config in this project (there is a missing space between `catch` and `(`)

daniloercoli:
Fixed the missing space. Yeah, we may want to add the checkstyle config. Maybe in another PR?

maxme:
> Maybe in another PR?

Yeah definitely not in that one, opened a ticket https://github.com/Automattic/Automattic-Tracks-Android/issues/31

maxme:
PR looks good, I'll publish a new version to bintray. We can update the dependency in wpandroid 7.2.

Issue #34, opened 2018-03-05
Set a max limit on the events queue size


daniloercoli:
We should set a max number of items the queue can contain. In case of long offline usage of the app, or in case of errors, the queue could grow up till to 10 thousand events and slow down the client network connection.

PR #35, opened 2018-03-05
Discard > 14 days old events, and set a limit of 10.000 items over the events queue

### LINKED ISSUE #33: Make sure the event is still valid before sending it to the server
On iOS we received some reports of a high number of duplicate events from a single client. Even though we fixed the underlying bug that caused those events a long time ago, they keep coming up, and we think the app might have them queued and it keeps trying to send them all.

Nothing similar reported on Android but better to change the function `TracksClient.isStillValid(Event)`, that now returns always `true`, by cheking the
- retry count 
- timestamp (expire after a while (1 week?))

Ref: https://github.com/Automattic/Automattic-Tracks-iOS/issues/68

daniloercoli:
This PR fixes #33 and #34 by making sure events are still valid before sending it to the server, and prevents possible floods of events sent to the server by setting a limit of 10.000 over the events queue.

CC @malinajirka and @khaykov  - Since you were the latest persons to work on the library since long time

### DIFF
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/TracksClient.java
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/TracksClient.java
@@ -111,6 +111,7 @@ public class TracksClient {
                         //2.  get the lock over the DB and write data
                         synchronized (mDbLock) {
                             // do not write events if the queue is already full
+                            // TODO: Remove older events and insert the new ones - See https://github.com/Automattic/Automattic-Tracks-Android/pull/35#discussion_r172458380
                             if (EventTable.getEventsCount(mContext) < DEFAULT_EVENTS_QUEUE_MAX_SIZE) {
                                 for (Event currentEvent : shadowCopyEventList) {
                                     EventTable.insertEvent(mContext, currentEvent);
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/TracksClient.java
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/TracksClient.java
@@ -34,7 +34,9 @@ public class TracksClient {
     protected static final String DEFAULT_USER_AGENT = "Nosara Client for Android" + "/" + LIB_VERSION;
     protected static final String NOSARA_REST_API_ENDPOINT_URL_V1_1 = "https://public-api.wordpress.com/rest/v1.1/";
     protected static final int DEFAULT_EVENTS_QUEUE_THRESHOLD = 9;
+    protected static final int DEFAULT_EVENTS_QUEUE_MAX_SIZE = 10000;
     protected static final int DEFAULT_EVENTS_QUEUE_TIMER_MS = 30000;
+    protected static final int DEFAULT_EVENT_MAX_AGE = 14 * 24 * 60 * 60 * 1000 ; // 14 days
 
     public static enum NosaraUserType {ANON, WPCOM, SIMPLENOTE}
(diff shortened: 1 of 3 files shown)


AutomatticTracks/src/main/java/com/automattic/android/tracks/TracksClient.java
```diff
@@ -108,8 +110,11 @@ public void run() {
                     if (shadowCopyEventList.size() > 0) {
                         //2.  get the lock over the DB and write data
                         synchronized (mDbLock) {
-                            for (Event currentEvent : shadowCopyEventList) {
-                                EventTable.insertEvent(mContext, currentEvent);
+                            // do not write events if the queue is already full
+                            if (EventTable.getEventsCount(mContext) < DEFAULT_EVENTS_QUEUE_MAX_SIZE) {
```
hypest: 👋 @daniloercoli . If I understand this correctly, the code will stop logging newer events if the queue is full. Have you perhaps dismissed the alternative of ejecting older events in favor of the newer ones? Especially with the added 14-day rule, keeping the older instead of the newer can get more events closer to that 14-day drop threshold.

daniloercoli: > Have you perhaps dismissed the alternative of ejecting older events in favor of the newer ones?

Not dismissed the idea at all, just wanted to keep the code simple for now, and dismissed new events only. It's not a common situation that the queue is full, in fact the issue was only reported on iOS since now. That was probably due to a bug in the iOS code, since we're sending events to the server as soon as there will be > 10 in the queue.

daniloercoli: Maybe we should add a comment there about this?

hypest: I think we can keep it as is then regarding the full queue policy.

Also, a comment can never hurt I guess so, 👍 for adding one.

Taking this opportunity to hash out a simple (perhaps) way out (but not pushing for it on this PR): wdyt about flushing the queue completely when it's full and a new event comes in? We could also inject a new event to track that we forcefully flushed the queue to accommodate for new events.

daniloercoli: > Taking this opportunity to hash out a simple (perhaps) way out (but not pushing for it on this PR): wdyt about flushing the queue completely when it's full and a new event comes in? We could also inject a new event to track that we forcefully flushed the queue to accommodate for new events.

This is already done in line 119 - `mDbLock.notifyAll();` - that is called regardless if new events are added or not. That code does trigger a chain of events that send all the tracked events already available in the queue to the server.

hypest: Oh, interesting. In that case I'm a bit puzzled: if we're flushing the queue one way or the other, why not let the new event be logged?

daniloercoli: We don't know if the network request, that is made immediately after, will finish without problems. In case of network errors (also when the device is offline) the flush is called, but does not send events to the server, and the queue is not cleared.

hypest: I see, thanks for expanding!

hypest (approved):
:shipit: !

PR #37, opened 2018-03-06
Tracks tablets and larger devices

### LINKED ISSUE #36: Add a property that keeps track if the device is a tablet
This is a hard one, since it's not clear when a device is a tablet or not.

We can calculate the diagonal of the screen in inches, and report it in a property as double. 
Also make assumptions, and add another property, boolean, like `is_large_screen` that is true when the screen diagonal is > 7 inches.

cc @iamthomasbishop

daniloercoli:
We need a simple way in our backend queries to identify if a device is a tablet or not. Actually we need a simple way to say if a device has a large screen or not. For now we say that a device has a large screen if the dimension is 7 inches or more. 

This PR basically calculates the real display size (diagonal), by using width/height pixels and display density. And use this new value to add 2 new properties `display_size` and `is_large_display` to events. Fixes #36.

cc @hypest and @iamthomasbishop 

Note: This PR also adds a utility method that removes all the events in the queue. (Not used anywhere atm).

### DIFF
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/DeviceInformation.java
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/DeviceInformation.java
@@ -7,6 +7,7 @@ import android.content.pm.ApplicationInfo;
 import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
 import android.content.pm.PackageManager.NameNotFoundException;
+import android.graphics.Point;
 import android.net.ConnectivityManager;
 import android.net.NetworkInfo;
 import android.os.Build;
@@ -37,6 +38,7 @@ import java.util.Locale;
 
 /* package */ class DeviceInformation {
     public static final String LOGTAG = "NosaraDeviceInformation";
+    private static final int DISPLAY_SIZE_LARGE_THRESHOLD = 7;
 
     private final Context mContext;
 
@@ -49,6 +51,8 @@ import java.util.Locale;
     private final Integer mAppVersionCode;
     private final Locale mLocale;
     private final String mDeviceLanguage;
+    private int mWidthPixels;
+    private int mHeightPixels;
 
     private final JSONObject mImmutableDeviceInfoJSON;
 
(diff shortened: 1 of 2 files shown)


xyu:
@bperson should we build something like this into the `w.js` Tracks client as well? Seems like categorizing based on screen size would be a simple way for us to approximate phone vs tablet vs laptop as opposed to trying to parse UAs and making sure that's updated.

bperson:
@xyu  We might as well directly use `devicewidth` / `deviceheight` ( which are already sent via `w.js` If I'm not mistaken ) and do the categorisation during the ETL? You're thinking about filling `browsertype` with that?

AutomatticTracks/src/main/java/com/automattic/android/tracks/DeviceInformation.java
```diff
@@ -113,10 +117,33 @@ public DeviceInformation(Context context) {
...
+                try {
+                    mWidthPixels = (int) Display.class.getMethod("getRawWidth").invoke(display);
+                    mHeightPixels = (int) Display.class.getMethod("getRawHeight").invoke(display);
+                } catch (Exception ignored) {
+                    Log.w(LOGTAG, "Unable to call getRawWidth/getRawHeight: " + ignored.getMessage());
+                }
+            } else if (Build.VERSION.SDK_INT >= 17) {
+                try {
+                    Point realSize = new Point();
+                    Display.class.getMethod("getRealSize", Point.class).invoke(display, realSize);
```
hypest: Wondering, why does this need reflection? Can't we just call `getRealSize` directly here that it's guarded by the `if (Build.VERSION.SDK_INT >= 17)`?

hypest:
> We might as well directly use devicewidth / deviceheight 

Hey @bperson , just to be sure it's not missed, let me note that the device's density will also be needed to compute the screen size in inches. We include that in `eventprops.device_info_display_density_dpi`.

AutomatticTracks/src/main/java/com/automattic/android/tracks/DeviceInformation.java
daniloercoli: Once you start use reflection you can't do without it :)
I've removed the unnecessary reflection calls.

hypest:
LGTM!

xyu:
> We might as well directly use devicewidth / deviceheight ( which are already sent via w.js If I'm not mistaken ) and do the categorisation during the ETL?

@bperson ah yeah this is a good point, I guess we already have this data in Tracks, I guess we can either categorize the screen size during the ETL or just install a UDF that takes both params and outputs a device category based on screen size. I guess during ETL would make queries faster but we would have to go back and rewrite old events.

> You're thinking about filling browsertype with that?

I guess maybe might make sense for a new column if we do this.

> the device's density will also be needed to compute the screen size in inches. We include that in eventprops.device_info_display_density_dpi

@hypest this is a really good point, looks like currently raw pixel counts are sent for the device height / width params:

https://github.com/Automattic/Automattic-Tracks-Android/blob/c2fdfa693019592ce1295b0b7207486dd4fb7b28/AutomatticTracks/src/main/java/com/automattic/android/tracks/DeviceInformation.java#L275-L281

Should we apply the same real point size logic there's here to those metrics as well?

@bperson I think `window.screen.width` for browsers already return logical pixels so we don't need to do any conversion there is this correct?

hypest:
> Should we apply the same real point size logic there's here to those metrics as well?

@xyu I guess on their own are fine as is but, not sure what's the historical reason we have them logged. I mean, maybe we need them as is (raw screen dimensions in real pixels) or perhaps we need to reconsider and convert them to device-independent pixels anyway. Any insight or information on the history here?

In the meantime yeah, for calculating screen size in the typical unit (inches) those will need conversion.

daniloercoli:
We've reported `width` and `height` because it seemed the standard way to report it.
Consider that even on desktop the real size of a pixel can vary. It depends on the monitor you have.

I think we can add 2 new methods that return `width/height` in inches. (if it's needed in our funnels). 
The diagonal was required to understand how many people use Tablets with our app. There is no standard way to get this info in Android, due to the fragmentation of the market.

PR #43, opened 2019-02-14
Upgrade Gradle and Plugin

shiki:
While working on wordpress-mobile/WordPress-Android#9253, I was blocked by not being able to build this project on Android Studio 3.3. 

## Sync

The Gradle sync fails with this error:

```
No toolchains found in the NDK toolchains folder for ABI with prefix: mips64el-linux-android
```

Upgrading Gradle and the Android plugin fixes the sync.

## Unresolved dependency

Added google and jcenter to repositories to fix this error when trying to run `./gradlew assemble`

```
Could not resolve all files for configuration ':AutomatticTracks:_internal_aapt2_binary'.
Could not find com.android.tools.build:aapt2:3.3.1-5013011.
```

It looks like aapt2 has been [relocated to `google()`](https://developer.android.com/studio/releases/#aapt2_gmaven). Interestingly, the build will not work without `jcenter()` too. I’m not sure why.

## Not handled by this PR

This error shows up when syncing but it does to seem to be too urgent to fix right now. 

```
WARNING: API 'variant.getJavaCompile()' is obsolete and has been replaced with 'variant.getJavaCompileProvider()'.
It will be removed at the end of 2019.
For more information, see https://d.android.com/r/tools/task-configuration-avoidance.
To determine what is calling variant.getJavaCompile(), use -Pandroid.debug.obsoleteApi=true on the command line to display a stack trace.
Affected Modules: AutomatticTracks
```

I also have no idea how to fix it. 🤷‍♂️

### DIFF
--- a/AutomatticTracks/build.gradle
+++ b/AutomatticTracks/build.gradle
@@ -4,7 +4,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.0.1'
+        classpath 'com.android.tools.build:gradle:3.3.1'
         classpath 'com.novoda:bintray-release:0.5.0'
     }
 }
@@ -22,7 +22,7 @@ dependencies {
 
 android {
     compileSdkVersion 26
-    buildToolsVersion "26.0.2"
+    buildToolsVersion "28.0.3"
 
     defaultConfig {
         versionName "1.1.5"
--- a/AutomatticTracks/build.gradle
+++ b/AutomatticTracks/build.gradle
@@ -24,7 +24,6 @@ dependencies {
 
 android {
     compileSdkVersion 26
-    buildToolsVersion "28.0.3"
 
     defaultConfig {
         versionName "1.1.5"
--- a/AutomatticTracks/build.gradle
+++ b/AutomatticTracks/build.gradle
@@ -16,7 +16,6 @@ apply plugin: 'com.novoda.bintray-release'
 repositories {
     google()
     jcenter()
-    mavenCentral()
 }
 
 dependencies {
--- a/AutomatticTracks/build.gradle
+++ b/AutomatticTracks/build.gradle
@@ -5,7 +5,7 @@ buildscript {
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:3.3.1'
-        classpath 'com.novoda:bintray-release:0.5.0'
+        classpath 'com.novoda:bintray-release:0.8.1'
     }
 }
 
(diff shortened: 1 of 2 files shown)


AutomatticTracks/build.gradle
```diff
@@ -22,7 +24,7 @@ dependencies {
 
 android {
     compileSdkVersion 26
-    buildToolsVersion "26.0.2"
+    buildToolsVersion "28.0.3"
```
malinajirka: According to the gradle plugin documentation it's no longer needed to explicitly specify android.buildToolsVersion. So I think we can just simply remove it. Wdyt?

`You no longer need to specify a version for the build tools (so, you can now remove the android.buildToolsVersion property). By default, the plugin automatically uses the minimum required build tools version for the version of Android plugin you're using.`
https://developer.android.com/studio/releases/gradle-plugin

malinajirka (commented):
Thanks @shiki!!

>WARNING: API 'variant.getJavaCompile()' is obsolete and has been replaced with 'variant.getJavaCompileProvider()'.

I've checked the stackTrace and it seems the 'com.novoda.bintray-release' plugin uses the old API. I wouldn't worry about it now.

>Added google and jcenter to repositories to fix this error when trying to run `./gradlew assemble`.

We might want to consider removing the `mavenCentral()` repo. AFAIU jcenter() is a superset of mavenCentral() (source: http://www.technotalkative.com/android-studio-migration-maven-central-jcenter/). Wdyt?


The `publishToMavenLocal` task doesn't work anymore. We'll need to update version of the `bintray-release` plugin. I've tried to update it to version`0.9` and it seems to work as expected (I just had to add "desc=xyz" to `publish` section in build.gradle.

Let me know what you think:). Thanks!

AutomatticTracks/build.gradle
shiki: @malinajirka 

> The `publishToMavenLocal` task doesn't work anymore.

Uh, yeah. Sorry about that. I fixed it in the other PR #44. I wanted to focus on the Gradle upgrade in this PR, in case I screwed things up. 

I'll cherry pick that to this PR. 

> According to the gradle plugin documentation it's no longer needed to explicitly specify android.buildToolsVersion. So I think we can just simply remove it. Wdyt?

Sure! I'll remove it.

shiki:
@malinajirka I removed `buildToolsVersion`, `mavenCentral()`, and upgraded Bintray. Ready for another review :)

malinajirka:
Awesome, thanks for the changes!! ;) LGTM

PR #44, opened 2019-02-14
Add is_online to event properties

shiki:
**Notice**: This PR depends on #43 to be merged first. 

This is part of wordpress-mobile/WordPress-Android#9253. The next step is linking the version of this PR (**1.1.6**) in WordPress-Android. 

 ## Testing

This can be tested by using this library directly in WordPress-Android.

1. Set this up as a local dependency in WordPress-Android as a local Maven artifact or a local project like in this [gist](https://gist.github.com/shiki/7d850e0a637b86e3738d5f3a030600eb). 
2. Run the WordPress-Android app. 
3. Play around with the app while offline so events are logged. 
4. Check the logged events of your WordPress account in Tracks Live View. Pick one event and validate that it has a `device_info_is_online` property. Please note that it takes a while for Tracks to display the logged events.

We probably won't be able to test this with Tracks Trends until the next day (p4qSXL-2Yz-p2) so the Tracks Live View is what we can test for now.

### DIFF
--- a/AutomatticTracks/build.gradle
+++ b/AutomatticTracks/build.gradle
@@ -27,7 +27,7 @@ android {
     buildToolsVersion "28.0.3"
 
     defaultConfig {
-        versionName "1.1.5"
+        versionName "1.1.6"
         minSdkVersion 14
         targetSdkVersion 25
     }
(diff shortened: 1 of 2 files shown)


malinajirka:
Thanks @shiki! Looks good;). I'll merge it and publish it to bintray as soon as #43 is merged.

PR #46, opened 2019-05-17
Add Sentry support

jkmassel:
This PR adds support for using Tracks as a wrapper around any crash logging service – in this case, Sentry.

If, in the future, apps needed to work with another crash logging provider, it'd be simple enough to swap out the implementation without disrupting any of the other applications.

Additionally, a lot of the low-level bookkeeping work is now represented here, ensuring that we don't have code duplication in many different applications. This will become more important as features such as log processing are added.

This PR is the counterpart to https://github.com/Automattic/Automattic-Tracks-iOS/pull/103

**Sample implementations:**
https://github.com/woocommerce/woocommerce-android/pull/1071

### DIFF
--- a/AutomatticTracks/build.gradle
+++ b/AutomatticTracks/build.gradle
@@ -19,6 +19,8 @@ repositories {
 }
 
 dependencies {
+    implementation 'io.sentry:sentry-android:1.7.16'
+    implementation 'org.jetbrains:annotations-java5:15.0'
 }
 
 android {
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.java
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.java
@@ -14,6 +14,7 @@ import java.util.Map;
 
 import io.sentry.Sentry;
 import io.sentry.SentryClient;
+import io.sentry.event.User;
 import io.sentry.android.AndroidSentryClientFactory;
 import io.sentry.connection.EventSendCallback;
 import io.sentry.event.Event;
@@ -74,48 +75,37 @@ public class CrashLogging {
         });
 
         mDataProvider = dataProvider;
-        applyUserTrackingPreferences();
+        setNeedsDataRefresh();
     }
 
     public static void crash() {
         throw new UnsupportedOperationException("This is a sample crash");
     }
 
-    public static boolean getUserHasOptedOut() {
-        return mDataProvider.getUserHasOptedOut();
+    public static void setNeedsDataRefresh() {
+        applyUserTracking();
+        applySentryContext();
     }
 
-    public static void setUserHasOptedOut(boolean userHasOptedOut) {
-        mDataProvider.setUserHasOptedOut(userHasOptedOut);
-        applyUserTrackingPreferences();
-    }
+    private static void applyUserTracking() {
+        sentry.getContext().clearUser();
 
-    private static void applyUserTrackingPreferences() {
-        if(!mDataProvider.getUserHasOptedOut()) {
-            enableUserTracking();
-        }
-        else{
-            disableUserTracking();
-        }
-    }
-
-    private static void enableUserTracking() {
         TracksUser tracksUser = mDataProvider.currentUser();
+
+        if (tracksUser == null) {
+            return;
+        }
 
         sentry.getContext().setUser(new UserBuilder()
                 .setEmail(tracksUser.getEmail())
                 .setUsername(tracksUser.getUsername())
                 .withData("userID", tracksUser.getUserID())
-                .build()
-        );
+                .setData(mDataProvider.userContext())
+                .build());
     }
 
-    private static void disableUserTracking() {
-        sentry.clearContext();
-    }
-
-    public static void setContext(Map<String, Object> context) {
-        sentry.setExtra(context);
+    private static void applySentryContext() {
+        sentry.setExtra(mDataProvider.applicationContext());
     }
 
     // Locale Helpers
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.java
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.java
@@ -1,0 +1,138 @@
+package com.automattic.android.tracks.CrashLogging;
+
+import android.content.Context;
+import android.util.Log;
+
+import com.automattic.android.tracks.BuildConfig;
+import com.automattic.android.tracks.TracksUser;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Locale;
+import java.util.Map;
+
+import io.sentry.Sentry;
+import io.sentry.SentryClient;
+import io.sentry.android.AndroidSentryClientFactory;
+import io.sentry.connection.EventSendCallback;
+import io.sentry.event.Event;
+import io.sentry.event.EventBuilder;
+import io.sentry.event.UserBuilder;
+import io.sentry.event.helper.ShouldSendEventCallback;
+import io.sentry.event.interfaces.ExceptionInterface;
+
+public class CrashLogging {
+
+    private static CrashLoggingDataProvider mDataProvider;
+    private static SentryClient sentry;
+
+    public static void start(Context context, CrashLoggingDataProvider dataProvider) {
+
+        sentry = Sentry.init(
+                dataProvider.sentryDSN(),
+                new AndroidSentryClientFactory(context)
+        );
+
+        // Apply Sentry Tags
+        sentry.setEnvironment(dataProvider.buildType());
+        sentry.setRelease(dataProvider.releaseName());
+        sentry.addTag("locale", getCurrentLanguage(dataProvider.locale()));
+
+        sentry.addShouldSendEventCallback(new ShouldSendEventCallback() {
+            @Override
+            public boolean shouldSend(Event event) {
+
+                if (BuildConfig.DEBUG) {
+                    return false;
+                }
+
+                return !mDataProvider.getUserHasOptedOut();
+            }
+        });
+
+        sentry.addEventSendCallback(new EventSendCallback() {
+            @Override
+            public void onFailure(Event event, Exception exception) {
+                Log.println(Log.WARN, "SENTRY", exception.getLocalizedMessage());
+            }
+
+            @Override
+            public void onSuccess(Event event) {
+                Log.println(Log.INFO, "SENTRY", event.getMessage());
+            }
+        });
+
+        mDataProvider = dataProvider;
+        applyUserTrackingPreferences();
+    }
+
+    public static void crash() {
+        throw new UnsupportedOperationException("This is a sample crash");
+    }
+
+    public static boolean getUserHasOptedOut() {
+        return mDataProvider.getUserHasOptedOut();
+    }
+
+    public static void setUserHasOptedOut(boolean userHasOptedOut) {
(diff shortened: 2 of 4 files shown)


AutomatticTracks/build.gradle
```diff
@@ -19,6 +19,8 @@ repositories {
 }
 
 dependencies {
+    implementation 'io.sentry:sentry-android:1.7.16'
+    implementation 'org.jetbrains:annotations-java5:15.0'
```
shiki, line 23: Thank you for adding those nullable/nonnull annotations. I find them really useful for expressing intentions in Java. 

In WPAndroid, it looks like we're using nullable/nonnull annotations from `com.android.support:support-annotations:28.0.0`. Does it make sense to make it consistent and use that in Tracks too?

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.java
```diff
@@ -0,0 +1,138 @@
...
+import io.sentry.event.helper.ShouldSendEventCallback;
+import io.sentry.event.interfaces.ExceptionInterface;
+
+public class CrashLogging {
+
+    private static CrashLoggingDataProvider mDataProvider;
+    private static SentryClient sentry;
+
+    public static void start(Context context, CrashLoggingDataProvider dataProvider) {
+
```
shiki, line 34: Is it safe for this method to be called multiple times? Or should we add a check here to prevent that from happening, just in case?

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLoggingDataProvider.java
```diff
@@ -0,0 +1,72 @@
...
+     * @see TracksUser
+     */
+    @Nullable TracksUser currentUser();
+
+    /**
+     * Provides the Crash Logging Service with information about the current system state.
+     *
+     * @see TracksUser
+     */
+    @NotNull  Map<String, Object> context();
```
shiki: I can't quite figure out how this is used. It is not used by `CrashLogging`.

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLoggingDataProvider.java
```diff
@@ -0,0 +1,72 @@
...
+     * Provides the Crash Logging Service with the Sentry DSN for this application.
+     * @return The Sentry DSN for this application.
+     */
+    String sentryDSN();
+
+    /**
+     * Provides the Crash Logging Service with information on whether the user has opted out
+     * of data collection.
+     *
+     * The implementation of this method should fetch this value from persistent storage, and should
```
shiki, line 22: > and should never call `CrashLogging.getUserHasOptedOut()`.

I feel like this should be enforced by `CrashLogging` instead. Perhaps make `CrashLogging.getUserHasOptedOut()` `private`? Or not have it at all? I'm thinking implementations (WP/Woo) can make use of their own utilities if they really need this information.

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLoggingDataProvider.java
```diff
@@ -0,0 +1,72 @@
...
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Locale;
+import java.util.Map;
+
+public interface CrashLoggingDataProvider {
+    /**
+     * Provides the Crash Logging Service with the Sentry DSN for this application.
```
shiki: Great comments in this class!

Would it make sense to use links instead? 

```java
* Provides {@link CrashLogging} with the Sentry DSN for this application.
```

In this way, it's clear which "service" we're referring to.

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLoggingDataProvider.java
```diff
@@ -0,0 +1,72 @@
...
+     *
+     * @return A value indicating whether or not the user has opted out of data collection.
+     */
+    boolean getUserHasOptedOut();
+
+    /**
+     * Allows persisting whether or not the user has opted out of data collection.
+     *
+     * The implementation of this method should write this value to persistent storage, and should
+     * never call `CrashLogging.setUserHasOptedOut()`.
```
shiki: Similar comment to the enforcement comment above. But this one is hard to avoid without making event listeners. Based on the Woo implementation though, it looks like we'll use this just for persistence. Should we probably rename it to `saveUserHasOptedOut()`.

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.java
```diff
@@ -0,0 +1,138 @@
...
+        return locale.getLanguage();
+    }
+
+    // Logging Helpers
+    public static void log(Throwable e) {
+        sentry.sendException(e);
+    }
+
+    public static void log(@NotNull Throwable e, Map<String, String> data) {
```
shiki: The `data` argument is not used. 🙂

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.java
```diff
@@ -0,0 +1,138 @@
...
+import io.sentry.SentryClient;
+import io.sentry.android.AndroidSentryClientFactory;
+import io.sentry.connection.EventSendCallback;
+import io.sentry.event.Event;
+import io.sentry.event.EventBuilder;
+import io.sentry.event.UserBuilder;
+import io.sentry.event.helper.ShouldSendEventCallback;
+import io.sentry.event.interfaces.ExceptionInterface;
+
+public class CrashLogging {
```
shiki, line 26: In the Woo implementation, we created a `CrashUtils` class which just passes its method contents like `log` and `logException` back to `CrashLogging`. I'm concerned that there would be some confusion about which class to use. Also, if someone accidentally uses `CrashLogging.logException()`, it will still work but without the custom logic implemented in `CrashUtils.logException()`. 

```
// this?
CrashLogging.logException()

// or this?
CrashUtils.logException()
```
My suggestion is to make `CrashLogging` not contain static methods at all and require it to be instantiated. Then, `CrashUtils` can instantiate it and configure it itself. 

```kt
object CrashUtils {
    private loggingService: CrashLogging
    fun initCrashLogging() {
        loggingService = CrashLogging()
        ...
    }
}
```

In this way, it will be a bit difficult to accidentally use `CrashLogging` itself and we would naturally rely on `CrashUtils` which has `static` `log()` and `logException()` methods. We can then safely add app-specific logic to `CrashUtils` without policing everyone to use it instead of `CrashLogging`. 

What do you think?

shiki (commented):
@jkmassel I'm sorry for getting to this a bit late. I did not receive an email notification when it was switched from _Draft_ to _Ready for Review_. 

I have some questions and comments for your consideration. 🙂

AutomatticTracks/build.gradle
jkmassel: It makes lots of sense!

Updated in 40afa31d06278c66411a2ef8c0d5848901dc26e5

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLoggingDataProvider.java
jkmassel: Nice! Addressed in 5133a08a825a9809da63b614d2aec445490c0971

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLoggingDataProvider.java
jkmassel: Addressed with a new architecture in 5133a08a825a9809da63b614d2aec445490c0971

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    public static void log(@NotNull Throwable e, Map<String, String> data) {
+
+        EventBuilder eventBuilder = new EventBuilder()
+                .withMessage(e.getMessage())
+                .withLevel(Event.Level.ERROR)
+                .withSentryInterface(new ExceptionInterface(e));
+
+        for (Map.Entry<String,String> entry : data.entrySet()) {
```
shiki: ⚠️ I guess `data` is `@Nullable`? Should we add a null check here?

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.java
```diff
@@ -0,0 +1,142 @@
...
+        applyUserTracking();
+        applySentryContext();
+    }
+
+    private static void applyUserTracking() {
+        sentry.getContext().clearUser();
+
+        TracksUser tracksUser = mDataProvider.currentUser();
+
+        if (tracksUser == null) {
```
shiki, line 95: ℹ️ Just wanted to point out that if `tracksUser` is `null`, the Sentry user is never cleared. It looks like Kotlin (`CrashUtils.kt` in WC) doesn't pick up the `@Nullable` annotation. 😞 We're always returning non-null in there.

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.java
```diff
@@ -0,0 +1,142 @@
...
+
+        mDataProvider = dataProvider;
+        setNeedsDataRefresh();
+    }
+
+    public static void crash() {
+        throw new UnsupportedOperationException("This is a sample crash");
+    }
+
+    public static void setNeedsDataRefresh() {
```
shiki, line 85: 🙌

shiki (approved):
I have a tiny important comment, marked with ⚠️. Aside from that, LGTM. 🎉 

Please see my [related approval message](https://github.com/woocommerce/woocommerce-android/pull/1071#pullrequestreview-244110696)  in https://github.com/woocommerce/woocommerce-android/pull/1071.

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.java
jkmassel: I think we're clearing it in line 91: https://github.com/Automattic/Automattic-Tracks-Android/blob/3b9a2b6ff6107a2cc0256243d5c1649c746ac26a/AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.java#L90-L105

But it's possible you're referring to something else and I didn't understand – let me know! :)

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.java
jkmassel: Nice catch! Better handling of this in 52d3e2340649794a4db4a94881840aa203dbf98c

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.java
shiki: Oh, you're right. I missed that line. 😄

PR #47, opened 2020-08-04
Use android annotations

aforcier:
Replaces the Jetbrains annotation package (introduced in https://github.com/Automattic/Automattic-Tracks-Android/pull/46 for nullability annotations) with the standard Android support library one.

I was working on an update to this library and noticed a dependency resolution error when using the existing `1.2` version in WPAndroid. It looks like we were importing version `15.0` of `org.jetbrains:annotations-java5` here and `13.0` somewhere else in WPAndroid's dependency tree. That could be solved by excluding on import, but I suspect there's no specific reason to use that library and we can instead just use the standard Android support library version of those annotations that won't cause any conflicts (I assume it was due to an auto-import).

@jkmassel if there was a specific reason for using that annotation library instead please let me know though.

I'll be opening another PR with my actual change to the library, I'll take care of bumping the version and pushing a new release to Bintray with that PR.

BTW I was surprised why WPAndroid isn't already using the latest version, aren't the Sentry changes needed there?

### To test
1. Check out this branch locally
2. Bump the `versionName` in `AutomatticTracks/build.gradle` to something like `1.2.1-rc-1`
3. Run `./gradlew clean assemble publishToMavenLocal` and check that it succeeds
4. In WPAndroid, open the root `build.gradle` file and add `mavenLocal()` to the top of the `repositories` block inside the `allprojects` block:
```groovy
allprojects {
    apply plugin: 'checkstyle'

    repositories {
        mavenLocal()
        google()
...
```
5. Still in WPAndroid, open `libs/analytics/WordPressAnalytics/build.gradle`, and bump `com.automattic:tracks` to the version you defined in step 2
6. Check that WPAndroid builds

### DIFF
--- a/AutomatticTracks/build.gradle
+++ b/AutomatticTracks/build.gradle
@@ -20,7 +20,7 @@ repositories {
 
 dependencies {
     implementation 'io.sentry:sentry-android:1.7.16'
-    implementation 'org.jetbrains:annotations-java5:15.0'
+    implementation 'com.android.support:support-annotations:28.0.0'
 }
 
 android {
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.java
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.java
@@ -1,13 +1,12 @@
 package com.automattic.android.tracks.CrashLogging;
 
 import android.content.Context;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 import android.util.Log;
 
 import com.automattic.android.tracks.BuildConfig;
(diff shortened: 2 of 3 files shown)


jkmassel (approved):
This is great, thanks!

> @jkmassel if there was a specific reason for using that annotation library instead please let me know though.

There was not – I just used the auto-suggested version when I was doing it – this is far better.

> BTW I was surprised why WPAndroid isn't already using the latest version, aren't the Sentry changes needed there?

They're not yet – WPAndroid still uses its own Sentry implementation. Syncing that up is coming – probably after encrypted logging lands?

aforcier:
> They're not yet – WPAndroid still uses its own Sentry implementation. Syncing that up is coming – probably after encrypted logging lands?

Ah gotcha 👍 Was just wondering. Thanks for the review!

PR #48, opened 2020-08-04
Fix width and height tracking

aforcier:
**~Note: While testing this PR I noticed that we seem to already be sending the data I needed as `_wd` and `_ht` properties, but those don't seem exposed in the Tracks frontend. I wonder if they're being filtered out automatically on the backend for some reason - I'm looking into this, this PR is on hold for now.~** (p4qSXL-3xm-p2)

_Once merged I'll push a new release to bintray and open a WPAndroid PR updating the version._

Adds two more properties to the default data we send with each tracks event: `device_info_display_height` and `device_info_display_width`.

**Update:** These are replacing the existing `_ht` and `_wd` params we'd already been sending. It turns out these were deprecated a few months ago and are ignored in the backend. (p4qSXL-3xm-p2)

This was prompted by ongoing work on the WP Stories feature. We're looking at standardizing exported slide sizes to a 9:16 ratio, and are assuming a relative rarity of WPAndroid devices with a wider than 9:16 ratio. Having some data on width/height will let us check our assumptions here.

Note that the width/height reported is [the _real_ width/height](https://github.com/Automattic/Automattic-Tracks-Android/blob/8c6b9b7dc3c8b63c10766ec7e2d9b61d09766cf7/AutomatticTracks/src/main/java/com/automattic/android/tracks/DeviceInformation.java#L146-L150), which is to say the real size of the display without subtracting any window decor. This is how we're also already calculating display size as hypotenuse, and should be appropriate given the stories view is a fullscreen experience, and this is inline with how other metrics websites report device screen size from what I can tell.

### To test
1. Check out this branch locally
2. Run `./gradlew clean assemble publishToMavenLocal` and check that it succeeds
3. In WPAndroid, open the root `build.gradle` file and add `mavenLocal()` to the top of the `repositories` block inside the `allprojects` block:
```groovy
allprojects {
    apply plugin: 'checkstyle'

    repositories {
        mavenLocal()
        google()
...
```
4. Still in WPAndroid, open `libs/analytics/WordPressAnalytics/build.gradle`, and bump `com.automattic:tracks` to `1.2.1-beta-1`
6. Check that WPAndroid builds
7. Hook up a network listener like Charles, run WPAndroid, and wait for a tracks event to be broadcast (https://public-api.wordpress.com/rest/v1.1/tracks/record) - backgrounding and foregrounding the app should trigger a call
8. Examine the data sent, you should see a `commonProps` object which should include height/width information:
```js
"commonProps": {
  ...
  "device_info_display_height": "2160",
  "device_info_display_width": "1080",
  ...
```
9. Also confirm that the `_ht` and `_wd` props are no longer included

### DIFF
--- a/AutomatticTracks/build.gradle
+++ b/AutomatticTracks/build.gradle
@@ -27,7 +27,7 @@ android {
     compileSdkVersion 26
 
     defaultConfig {
-        versionName "1.2"
+        versionName "1.2.1-beta-1"
         minSdkVersion 14
         targetSdkVersion 25
     }
--- a/AutomatticTracks/build.gradle
+++ b/AutomatticTracks/build.gradle
@@ -27,7 +27,7 @@ android {
     compileSdkVersion 26
 
     defaultConfig {
-        versionName "1.2.1-beta-1"
+        versionName "1.2.1-beta-2"
         minSdkVersion 14
         targetSdkVersion 25
     }
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/DeviceInformation.java
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/DeviceInformation.java
@@ -195,7 +195,9 @@ import java.util.Locale;
         }
         try {
             // Width and height depend on device orientation - to be consistent, always
-            // report the shorter dimension as the width
+            // report the shorter dimension as the width.
+            // These values represent the 'real' dimensions of the screen, ignoring navigation
+            // and window decoration.
             if (mHeightPixels > mWidthPixels) {
                 mImmutableDeviceInfoJSON.put("display_height", mHeightPixels);
                 mImmutableDeviceInfoJSON.put("display_width", mWidthPixels);
@@ -205,6 +207,21 @@ import java.util.Locale;
             }
         } catch (final JSONException e) {
             Log.e(LOGTAG, "Exception writing display_height and width value in JSON object", e);
+        }
+        try {
+            // Width and height depend on device orientation - to be consistent, always
+            // report the shorter dimension as the width.
+            // These values represent the usable dimensions of the screen - whatever is left after
+            // navigation and window decoration.
+            if (mDisplayMetrics.heightPixels > mDisplayMetrics.widthPixels) {
+                mImmutableDeviceInfoJSON.put("display_usable_height", mDisplayMetrics.heightPixels);
+                mImmutableDeviceInfoJSON.put("display_usable_width", mDisplayMetrics.widthPixels);
+            } else {
+                mImmutableDeviceInfoJSON.put("display_usable_height", mDisplayMetrics.widthPixels);
+                mImmutableDeviceInfoJSON.put("display_usable_width", mDisplayMetrics.heightPixels);
+            }
+        } catch (final JSONException e) {
+            Log.e(LOGTAG, "Exception writing display_usable_height and width value in JSON object", e);
         }
         try {
             mImmutableDeviceInfoJSON.put("bluetooth_version", getBluetoothVersion());
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/DeviceInformation.java
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/DeviceInformation.java
@@ -182,8 +182,8 @@ import java.util.Locale;
             int densityDpi = getDisplayMetrics().densityDpi;
             mImmutableDeviceInfoJSON.put("display_density_dpi", densityDpi);
             if (densityDpi > 0) {
-                double height = getDeviceHeightPixels() / (double) densityDpi;
-                double width = getDeviceWidthPixels() / (double) densityDpi;
+                double height = mHeightPixels / (double) densityDpi;
+                double width = mWidthPixels / (double) densityDpi;
                 double size = Math.hypot(width, height);
                 // Format it now
                 size = Math.round(size * 10d) / 10d;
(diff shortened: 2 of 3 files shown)


aforcier:
Ready for review now @mzorz !

mzorz:
Thanks for putting this together @aforcier! 
Followed the instructions to test and confirmed it works :+1:

Screen Shot 2020-08-06 at 09 08 39


I was wondering how the height and width is collected and looking into [this portion of code here](https://github.com/Automattic/Automattic-Tracks-Android/pull/48/files#diff-1718aa6474e6bc8491113e238aa9e95aR134-L154); a few things to note:

- we could drop the `if (Build.VERSION.SDK_INT >= 14 && Build.VERSION.SDK_INT < 17) {` section since we're only supporting SDK 21 and later now on WordPressAndroid  - I would understand we may want to leave this here if this library was used elsewhere / by others but I believe this is only used for our own Tracks tool.

```
Note that the width/height reported is the real width/height, which is to say the real size of the display without subtracting any window decor. This is how we're also already calculating display size as hypotenuse, and should be appropriate given the stories view is a fullscreen experience, and this is inline with how other metrics websites report device screen size from what I can tell.
```
- Good point in using  `getRealSize()` instead of informing the Display size as reported by `mDisplayMetrics.widthPixels;` and 
`mDisplayMetrics.heightPixels;`. I tested this on the side and seems to work well (it was working before ofc, but just to make sure we get what we want). 

Related to this, you mention the following
> [...] and should be appropriate given the stories view is a fullscreen experience [...]

In this I'm wondering, wouldn't we like to take the opportunity and  _also_ want to track the measurements as given by `mDisplayMetrics`'s `widthPixels` and `heightPixels`? I ask because even if the phone screen aspect ratio is 9:16, we're still needing to show the navigation controls and that borrows some space (see the "important note" part in [this PR's description](https://github.com/Automattic/stories-android/pull/479)), and this breaks the 9:16 relationship which is what we are trying to validate as well (how much space is taken, to inform the decision on how much cropping we'll be doing and as such, how much information / interactable screen area could be lost)?

- I understand we may want to be consistent with previous data that we may have stored (unless it's wiped now?), wondering whether we should also name the informed attribute the same as per what we had before (that is `_ht` and `_wd` params), and lift the filter server-side (just an open question; I think we should be fine to have the more consistent `device_info_` prefix anyway).

- if not, and then  `_ht` and `_wd` are not used anymore anyway (because we're maybe filtering them server-side on arrival) then we may want to get rid of the public methods `getDeviceWidthPixels()` and `getDeviceHeightPixels()` in DeviceInformation if not used anymore, and change the reference to the class members `mHeightPixels` and `mWidthPixels` [here](https://github.com/Automattic/Automattic-Tracks-Android/pull/48/files#diff-1718aa6474e6bc8491113e238aa9e95aL185-L186).

Let me know your thoughts 🙇

aforcier:
Thanks @mzorz ! A few answers:

> we could drop the if (Build.VERSION.SDK_INT >= 14 && Build.VERSION.SDK_INT < 17) { section since we're only supporting SDK 21 and later now on WordPressAndroid - I would understand we may want to leave this here if this library was used elsewhere / by others but I believe this is only used for our own Tracks tool.

🤷‍♂️ it's only used by our own apps, but our usual approach to bumping the minSdk for things is to weigh the benefits. I think it's unlikely, but we might one day want to use this library for an app with older sdk support, and this doesn't seem like a sufficient reason to bump the minSdk to a pretty basic library 🤔 

> In this I'm wondering, wouldn't we like to take the opportunity and also want to track the measurements as given by mDisplayMetrics's widthPixels and heightPixels? I ask because even if the phone screen aspect ratio is 9:16, we're still needing to show the navigation controls and that borrows some space (see the "important note" part in this PR's description), and this breaks the 9:16 relationship which is what we are trying to validate as well (how much space is taken, to inform the decision on how much cropping we'll be doing and as such, how much information / interactable screen area could be lost)?

Sure - I added `device_info_display_usable_height` and `device_info_display_usable_width` in 
93fb421  👍 

As I suggested on https://github.com/Automattic/stories-android/pull/479 we may not end up needing those for our purposes if we can hide the navigation bar for stories. Also I think it might be more appropriate to track this sort of thing attached to a specific story event than a general param. I made [a note](https://github.com/Automattic/stories-android/issues/476#issuecomment-670231502) to make sure we add our own cropping tracking, and we can eventually remove the params I just added.

> I understand we may want to be consistent with previous data that we may have stored (unless it's wiped now?), wondering whether we should also name the informed attribute the same as per what we had before (that is _ht and _wd params), and lift the filter server-side (just an open question; I think we should be fine to have the more consistent device_info_ prefix anyway).
> if not, and then _ht and _wd are not used anymore anyway (because we're maybe filtering them server-side on arrival) then we may want to get rid of the public methods getDeviceWidthPixels() and getDeviceHeightPixels() in DeviceInformation if not used anymore, and change the reference to the class members mHeightPixels and mWidthPixels here.

There's no data to be consistent with, it's all gone (see p4qSXL-3xm-p2). Underscore-prefixed params are generally 'reserved' ones with special meaning for tracks, as that one is deprecated we should definitely just not use it anymore.

I removed the two unneeded public methods in 98254e6. 👍 

Ready for another pass @mzorz !

mzorz:
> 🤷‍♂️ it's only used by our own apps, but our usual approach to bumping the minSdk for things is to weigh the benefits. I think it's unlikely, but we might one day want to use this library for an app with older sdk support, and this doesn't seem like a sufficient reason to bump the minSdk to a pretty basic library 🤔

Good points, makes sense

> As I suggested on Automattic/stories-android#479 we may not end up needing those for our purposes if we can hide the navigation bar for stories. Also I think it might be more appropriate to track this sort of thing attached to a specific story event than a general param. I made a note to make sure we add our own cropping tracking, and we can eventually remove the params I just added.

Brilliant, I think this works 👍 

> There's no data to be consistent with

😢 

> I removed the two unneeded public methods in 98254e6. 👍

Thank you!

mzorz (approved):
Reviewed, LGTM

The Lord of Albatross says:
wp_story1596759459923_0

PR #51, opened 2021-02-24
Bump Sentry SDK to 4.3.0

### LINKED ISSUE #50: Bump Sentry SDK to 4.3.0
It's necessary to bump the SDK to at least `2.x.x` to have access to `SentryOptions#setBeforeSend`, necessary to go with encrypted logs used in `WP/WCAndroid`. But at this point I can't see any reason against bumping to the newest possible version of SDK which is `4.3.0`

wzieba:
This PR bumps Sentry SDK to `4.3.0` and migrates crash logging code to Kotlin in order to achieve more straightforward integration with clients.

Resolves #50


AutomatticTracks/build.gradle
```diff
@@ -4,37 +4,48 @@ buildscript {
...
-    compileSdkVersion 26
+    compileSdkVersion 30
 
     defaultConfig {
-        versionName "1.2.1"
+        versionName "${versionName}"
         minSdkVersion 14
-        targetSdkVersion 25
+
+        buildConfigField "String", "VERSION_NAME", "${versionName}"
```
wzieba: [From AGP release notes](https://developer.android.com/studio/releases/gradle-plugin#version_properties_removed_from_buildconfig_class_in_library_projects):
> Version properties removed from BuildConfig class in library projects
For library projects only, the BuildConfig.VERSION_NAME and BuildConfig.VERSION_CODE properties have been removed from the generated BuildConfig class because these static values did not reflect the final values of the application’s version code and name, and were therefore misleading. Additionally, these values were discarded during manifest merging.

I think it's something we should handle later, I don't know how yet - probably when migrating of JCenter?

I've decided to artificially add keep previous implementation.

AutomatticTracks/src/test/java/com/automattic/android/tracks/CrashLogging/CrashLoggingTest.kt
```diff
@@ -0,0 +1,78 @@
...
+    @Test(expected = UnsupportedOperationException::class)
+    fun shouldThrowExceptionOnRequest() {
+        CrashLogging.crash()
+    }
+
+    @Test
+    fun `should call for required arguments during initialisation`() {
+        initialize()
+
+        verify(dataProvider, times(1)).sentryDSN()
```
wzieba: I didn't want to use `PowerMock` to mock static `Sentry` or `SentryAndroid` so I've decided to spy on our `CrashLoggingDataProvider` to make sure methods with data are requested (so we can assume that are assigned).

AutomatticTracks/build.gradle
```diff
@@ -4,37 +4,48 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.3.1'
+        classpath 'com.android.tools.build:gradle:4.1.2'
```
oguzkocer: Although this wouldn't make a difference right now, I think it'd be good to use the same version as our other projects as otherwise composite builds will not work. We plan to upgrade all projects to latest Gradle version and gradle plugin, but we also plan to have this library available as a composite build and I am not sure which will happen first. Just to make things easier, I suggest using version `4.0.2` [to match WCAndroid](https://github.com/woocommerce/woocommerce-android/blob/develop/build.gradle#L10). We use `4.0.1` in WPAndroid, but I think we can update that to `4.0.2` without any changes.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Thu Feb 14 14:07:57 MST 2019
+#Wed Feb 24 11:46:10 CET 2021
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.5-bin.zip
```
oguzkocer: Similar to my other comment about the Android Gradle Plugin, I think we should match the version we use in our other apps which is `6.1.1`. Also, we should use the following command to achieve this, so all the relevant files will get updated:

```
./gradlew wrapper --gradle-version 6.1.1 --distribution-type all
```

For example, `gradlew` and `gradlew.bat` files are not currently updated in the PR.

We also prefer the `all` version over `bin` one. Admittedly, I don't know if we actually utilize the source files and the documentation, but I think it's still the better choice.

gradle.properties
```diff
@@ -0,0 +1,2 @@
+android.enableJetifier=true
+android.useAndroidX=true
```
oguzkocer: It's typically not a good idea to have `gradle.properties` checked into git since conventionally it's used to store sensitive information. In WPAndroid and WCAndroid, we have this file in `.gitignore` [#](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/.gitignore#L52), [#](https://github.com/woocommerce/woocommerce-android/blob/develop/.gitignore#L24) and instead utilize a `gradle.properties-example` file. For the CI, we use a step to copy the example file, or decrypt one with actual values depending on the task.

We unfortunately have some libraries that does this exact same thing, but we should be changing those as well. It's not harmful if everyone is careful not to push anything sensitive, but there is no reason to risk it.

_P.S: I think it was a poor decision to require these properties in the `gradle.properties` file 🤷‍♂️_

AutomatticTracks/build.gradle
```diff
@@ -4,37 +4,48 @@ buildscript {
...
+    implementation 'io.sentry:sentry-android:4.1.0'
+    implementation "androidx.annotation:annotation:1.1.0"
+    testImplementation "junit:junit:4.13"
+    testImplementation "org.mockito:mockito-core:3.3.3"
+    testImplementation 'org.robolectric:robolectric:4.4'
+    testImplementation 'androidx.test:core:1.3.0'
+}
+
+project.ext {
+    versionName = "1.2.1"
```
oguzkocer: I think this is not a great name because it matches the keyword android DSL expects. And.. I don't think it actually works. Currently, `VERSION_NAME` field gets assigned to `null`.

I don't think we need to make it this complicated, the following should work as expected:

```
android {
    compileSdkVersion 30

    defaultConfig {
        versionName "1.2.1"
        minSdkVersion 14

        buildConfigField "String", "VERSION_NAME", "\"$versionName\""
    }

    compileOptions {
        sourceCompatibility JavaVersion.VERSION_1_8
        targetCompatibility JavaVersion.VERSION_1_8
    }
}
```

AutomatticTracks/build.gradle
oguzkocer: I think we could retrieve this property dynamically from the package info. I have seen some stackoverflow threads, but I haven't had a chance to try it yet. Maybe it'd be worth looking into as we may miss this in future refactors. At the very least, I think we should add a TODO for it.

I agree with their decision to remove it, because a lot of the time, the version name will not actually match the value in the build file. This is because publishing a new version would require the build file to change and that's not fun to do in the CI or ask developers to make sure of. It's just easier (better?) to rely on information from git. (such as tags)

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/BeforeSendCallback.kt
```diff
@@ -0,0 +1,17 @@
...
+        private val dataProvider: CrashLoggingDataProvider,
+) : SentryOptions.BeforeSendCallback {
+    override fun execute(event: SentryEvent, hint: Any?): SentryEvent? {
+        return if (debug || dataProvider.userHasOptedOut) {
+            null
+        } else {
+            event
+        }
+    }
+}
```
oguzkocer: I think it'd be better to expose a callback to the clients so they can be the ones who decide if a Sentry event should be sent. We also do [a lot of configuration in the WPAndroid in this callback](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/WordPress/src/main/java/org/wordpress/android/util/CrashLogging.kt#L32-L82), which I don't think this interface currently allows to do. Actually, can we please make this an interface if we end up keeping it?

I _think_ we may have to expose the SentryEvent, at least partially to the client to be able to do all that configuration. I haven't given this any thought yet, but let me know if you need any help with it.

P.S: I think maybe this is out of the scope for this PR and you may be planning to make these types of changes in a future PR. If that's the case, you can ignore it for this PR, but maybe it'll be helpful for the future one.

oguzkocer (commented):
@wzieba I've left some comments, especially about tooling. However, I am reluctant to leave feedback about the actual implementation, because I can see that a lot of these changes are due to Java - Kotlin conversion.

I like that you're adding Kotlin support, but I think it'd have been much better to do that in a separate PR with no functional changes to the code. That way, it'd have been much easier to focus on your changes.

Furthermore, it looks like you forced pushed this branch a couple times after opening the PR. That's something I strongly discourage mainly because we can't easily see what changed since it's opened. (reviewed) In this case I haven't had started reviewing, so no harm done, but please consider adding new commits if there is anything you need to change.

On the topic of commits, this whole PR is just a single commit which makes it difficult to see how the PR has evolved. If for example you have added commits for the auto Kotlin conversion, it'd have been much easier to ignore them. We have several p2 posts that talks about writing good commit messages, it may be valuable to visit them.

---

I think some of this might be more than you asked for. I am sorry to have brought up several non-implementation topics, but I still hope that you'll find them useful.

On a personal note, I am wrapping up a very high priority project and have very limited availability for a few days. I am doing my best to carve out some time, but to speed things along, please feel free to work with other product developers for the implementation details. I can still get involved on the high level changes, but may not be able to do another line by line review until next week.

gradle/wrapper/gradle-wrapper.properties
wzieba: Thanks, totally forgot about wrappers 👍 

I get your point with having `all` over `bin`, I don't have any strong opinion on that especially that size difference is not that big as I thought it can be:
```
$ ls -la gradle-6.1.1-*
.rw-r--r--@ 138M wzieba 26 Feb  8:22 gradle-6.1.1-all.zip
.rw-r--r--@  97M wzieba 26 Feb  8:22 gradle-6.1.1-bin.zip
```
I'll change it to `all`

gradle.properties
wzieba: Thanks, TIL 
> conventionally it's used to store sensitive information

I've never considered this file as a storage for any sensitive information, but if anyone does, I'm ok with having this in `.gitignore` and use strategy from `WPAndroid` or `WCAndroid` you mention.

AutomatticTracks/build.gradle
wzieba: Right, thanks for pointing it. It indeed will be null. I'll take care of this change in next PRs.

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/BeforeSendCallback.kt
wzieba: This class was meant to have `internal` access modifier - I didn't plan to expose it to client, I've extracted it mostly to split responsibilities and unit test it.

> I think we may have to expose the SentryEvent, at least partially to the client to be able to do all that configuration.

That's probably true, your link to WPAndroid illustrates it perfectly but maybe we could abstract it instead of coupling `WPAndroid` with `Sentry` - I'll think of it in future PR.

> but maybe it'll be helpful for the future one.

Sure it is, thanks!

wzieba:
Thank you a lot for a review @oguzkocer , let me explain my reasoning on high level things you mention:

> I like that you're adding Kotlin support, but I think it'd have been much better to do that in a separate PR with no functional changes to the code

Sure thing, I'll do this in separate PR

> Furthermore, it looks like you forced pushed this branch a couple times after opening the PR. That's something I strongly discourage mainly because we can't easily see what changed since it's opened. (reviewed)

I'm happy that no harm was done, changes I forced push were a minor ones like adding next line at the end of a file or changing lib version which I eventually reverted.

I amend commit because merge strategy used in this and other projects is `merge commit` which will end up with merging all of the commits in PR into target branch `develop`. I think many non important commits like "fixed lint" can shadow important commits and create messy git log, especially on target branch with a lot of commits.

But that's just an opinion. I understand the strategy with not amending commits and using `merge commit` works here and other projects so I'm ok with stop amending the commits.

> On the topic of commits, this whole PR is just a single commit which makes it difficult to see how the PR has evolved.

True, thanks for pointing it out. In this case it was difficult not to do this to not block myself with e.g. Kotlin support but I understand your point. I'll split this PR into smaller ones and try to keep commit history as a more divided one.

> I think some of this might be more than you asked for. I am sorry to have brought up several non-implementation topics, but I still hope that you'll find them useful.

Yes, I find them very useful. Thank you a lot for that, it'll definitely help me to work better not only on this, but other projects too.

> On a personal note, I am wrapping up a very high priority project and have very limited availability for a few days. (...) I can still get involved on the high level changes.

Sure, if it's ok I'll assign to you some reviews of tools updates, feel free to review them when you have time. I'll make sure that those won't block feature parts I'm working on. Thanks for heads up!

gradle.properties
oguzkocer: You may want to check out the `gradle.properties` file in your WPAndroid and WCAndroid folders :) You are probably using our configuration tool to fetch/update the secrets, so you may not have added anything sensitive yet, but they are certainly there.

Whether that's the perfect place for them is questionable to me as well, but it's the convention and currently what we use. On one hand, it makes it a lot easier to access the properties since Gradle will automatically load them from the `gradle.properties` file, but on the other, it might be better to have a separate file for those to avoid this type of confusion. 🤷‍♂️

oguzkocer:
> I amend commit because merge strategy used in this and other projects is merge commit which will end up with merging all of the commits in PR into target branch develop. I think many non important commits like "fixed lint" can shadow important commits and create messy git log, especially on target branch with a lot of commits.

I understand where you're coming from. I amend, rebase, reword commits very often, but I just do it before I open the PR. I like to review my own PR before creating it, which seems to help with these kinds of issues. Anyway, it's good to limit those extra commits as much as we can, but it shouldn't overshadow more important principles, so I suggest not worrying too much about them.

> Sure, if it's ok I'll assign to you some reviews of tools updates, feel free to review them when you have time. I'll make sure that those won't block feature parts I'm working on. Thanks for heads up!

Of course and I'll do my best to get to them as quickly as I can!

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/BeforeSendCallback.kt
oguzkocer: OK, it's a bit difficult to give any meaningful feedback on this before we get a sense of what you have in mind, so I'll wait to follow up on this.

AutomatticTracks/build.gradle
wzieba: In #54 we bumped AGP to `4.0.2` instead, so this field is still available. Let's go back to this when updating AGP to `4.1.x`

AutomatticTracks/build.gradle
wzieba: Fixed in #54

gradle/wrapper/gradle-wrapper.properties
wzieba: Fixed in #54

AutomatticTracks/build.gradle
wzieba: The same as here https://github.com/Automattic/Automattic-Tracks-Android/pull/51#discussion_r587282929 - let's postpone it to AGP `4.1.x` update

gradle.properties
wzieba: Sure, that's good idea to follow the same convention as in other projects. I've addressed this in #62

AutomatticTracks/src/main/AndroidManifest.xml
```diff
@@ -1,4 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest package="com.automattic.android.tracks">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="com.automattic.android.tracks">
 
+    <application>
+        <meta-data
+            android:name="io.sentry.auto-init"
+            android:value="false" />
+    </application>
```
wzieba, line 12: This will be eventually merged into the client's manifest:
Screenshot 2021-03-04 at 14 08 51

This disables automated Sentry initialisation. We do this in CrashLogging#start with `dsn` provided by the client.
Reference: https://docs.sentry.io/platforms/android/migration/#manual-installation

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.kt
```diff
@@ -0,0 +1,101 @@
...
+import io.sentry.Sentry
+import io.sentry.SentryEvent
+import io.sentry.SentryLevel
+import io.sentry.SentryOptions
+import io.sentry.android.core.SentryAndroid
+import io.sentry.protocol.Message
+import io.sentry.protocol.User
+import java.util.*
+
+object CrashLogging {
```
wzieba: I've decided to propose rewriting `CrashLogging` to Kotlin to have a better use of Sentry's ["Unified API"](https://develop.sentry.dev/sdk/unified-api/)

So e.g. instead of
```java
SentryEvent event = new SentryEvent();
Message message = new Message();
message = throwable.message;
event.message = message;
event.level = SentryLevel.ERROR;
event.setExtras(data);
Sentry.captureEvent(event);
```

```kotlin
val event = SentryEvent().apply {
    message = Message().apply {
        message = throwable.message
    }
    level = SentryLevel.ERROR
    setExtras(data)
}
Sentry.captureEvent(event)
```

the same applies to e.g. `User`. IMO it's quite more readable.

Also 2 of 3 clients of `Tracks` are mostly in Kotlin so it's easier to assert nullability and mutability

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLoggingDataProvider.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * Provides the [CrashLogging] with information about the current application state.
+     */
+    val applicationContext: Map<String, String?>
+
+    /**
+     * Provides the [CrashLogging] with information about the current application state.
+     */
+    val userContext: Map<String, String?>
```
wzieba: I've changed types from `Map<String, Object>` (it was java so `Object` was nullable) to `Map<String, String?>`

The reason is that Sentry dropped method `Sentry.setExtra(Map<String, Object>)`. New one is `Sentry.setExtra(String, String)`. We could leave the old interface and call `toString` on `Object` but I didn't want to hide this from client. IMO it's better that client will choose what's exactly extra value it adds.

AutomatticTracks/src/androidTest/java/com/automattic/android/tracks/CrashLogging/SendEventsToCrashLoggerService.kt
```diff
@@ -0,0 +1,61 @@
...
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.platform.app.InstrumentationRegistry
+import com.automattic.android.tracks.TracksUser
+import com.automattic.android.tracks.fakes.FakeDataProvider
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class SendEventsToCrashLoggerService {
```
wzieba: The idea for this "service" is to have some mechanism for frequent end-to-end check of the Sentry integration. It's not a test, even if it's in `androidTest` directory.

This way we're to run a lot of tests on test Sentry project. I believe this will be helpful also for future work.

## Demo


https://user-images.githubusercontent.com/5845095/111218209-b91b4600-85d6-11eb-92e4-8760d9a23273.mp4

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/BeforeSendCallback.kt
wzieba: I've ended up with not extracting this callback for now. I'll follow up to your comments in next PRs.

wzieba:
@oguzkocer @jkmassel just to confirm - this PR is ready for re-review. Please take a look when you have time. Thanks!

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.kt
```diff
@@ -0,0 +1,104 @@
...
+        SentryAndroid.init(context) { options ->
+            options.apply {
+                dsn = dataProvider.sentryDSN
+                environment = dataProvider.buildType
+                release = dataProvider.releaseName
+                setDebug(BuildConfig.DEBUG)
+                setLogger(if (BuildConfig.DEBUG) SystemOutLogger() else NoOpLogger.getInstance())
+                setTag("locale", getCurrentLanguage(dataProvider.locale))
+                beforeSend = SentryOptions.BeforeSendCallback { event, _ ->
+                    return@BeforeSendCallback if (dataProvider.userHasOptedOut) {
```
wzieba: Here's also change I'd like to border. Previously we didn't send event if library was in `debug` variant but I'd like to leave this decision only to client. There are cases when we want to send event in `debug` (testing)

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.kt
```diff
@@ -0,0 +1,109 @@
+package com.automattic.android.tracks.crashlogging
```
oguzkocer: For several files, I see that Android Studio gives this warning: `Package directive doesn't match file location `. I think the folder names need to be changed. Could you address this?

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLoggingDataProvider.kt
```diff
@@ -0,0 +1,59 @@
...
+     * @return A `TracksUser` object containing data about the current user.
+     *
+     * @see TracksUser
+     */
+    val currentUser: TracksUser?
+
+    /**
+     * Provides the [CrashLogging] with information about the current application state.
+     */
+    val applicationContext: Map<String, String?>
```
oguzkocer: Do we need the value to be nullable? Does a key without a value meaningful in this context?

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.kt
```diff
@@ -0,0 +1,109 @@
...
+import io.sentry.SentryOptions
+import io.sentry.SystemOutLogger
+import io.sentry.android.core.SentryAndroid
+import io.sentry.protocol.Message
+import io.sentry.protocol.User
+import java.util.Locale
+
+object CrashLogging {
+
+    private const val EMPTY_VALUE = ""
```
oguzkocer: Is there any value in having a constant with just an empty string value? I understand the inclination to use constants in general, but an empty string feels like an overkill to me. What do you think?

If you really want to keep it, can we please rename it to `EMPTY_STRING` instead? `EMPTY_VALUE` is less meaningful in this case than a simple `""` since it can refer to anything.

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.kt
```diff
@@ -0,0 +1,109 @@
...
+            }
+        }
+        this.dataProvider = dataProvider
+        setNeedsDataRefresh()
+    }
+
+    @JvmStatic
+    fun crash() {
+        throw UnsupportedOperationException("This is a sample crash")
+    }
```
oguzkocer: I see that this was in the original Java implementation as well, but it might be worth considering to remove it. I don't think it provides much context or value since we can simply throw an exception wherever we need it. If we did that, we could probably use a better message for it since we'd have more context about why we are crashing on purpose. What do you think?

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.kt
```diff
@@ -0,0 +1,109 @@
...
+            Sentry.setExtra(entry.key, entry.value ?: EMPTY_VALUE)
+        }
+    }
+
+    // Locale Helpers
+    private fun getCurrentLanguage(locale: Locale?): String {
+        return if (locale == null) {
+            "unknown"
+        } else locale.language
+    }
```
oguzkocer: This seems to be used in a single place. Would it make sense to simply inline this instead?

```
locale?.language ?: "unknown"
```

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLoggingDataProvider.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * Provides the [CrashLogging] with information about the current application state.
+     */
+    val userContext: Map<String, String?>
```
oguzkocer: The following question about `applicationContext` is applicable here as well:

> Do we need the value to be nullable? Does a key without a value meaningful in this context?

AutomatticTracks/src/test/java/com/automattic/android/tracks/CrashLogging/CrashLoggingTest.kt
```diff
@@ -0,0 +1,70 @@
...
+
+        user = if (currentUser != null) spy(currentUser) else null
+
+        dataProvider = spy(FakeDataProvider(currentUser = user))
+
+        CrashLogging.start(
+            context = RuntimeEnvironment.systemContext,
+            dataProvider = dataProvider
+        )
+    }
```
oguzkocer: I am not a big fan of using global properties for tests because it leads to the possibility of the value being changed. Since we already call the `initialize` method from almost all the tests, would you consider a solution such as this:

```
private fun setup(currentUser: TracksUser? = null): Pair<CrashLoggingDataProvider, TracksUser?> {
    val user = if (currentUser != null) spy(currentUser) else null

    val dataProvider = spy(FakeDataProvider(currentUser = user))

    CrashLogging.start(
        context = RuntimeEnvironment.systemContext,
        dataProvider = dataProvider
    )
    return Pair(dataProvider, user)
}
```

With this ^, we can simply change the `initialize()` calls to `val (dataProvider, _) = setup()` and everything else should just work. (except for the specific user case obviously)

AutomatticTracks/src/test/java/com/automattic/android/tracks/CrashLogging/CrashLoggingTest.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    @Test(expected = UnsupportedOperationException::class)
+    fun shouldThrowExceptionOnRequest() {
+        CrashLogging.crash()
+    }
```
oguzkocer: Hmm.. I am a little confused. This seems to just call a function that throws an exception. The `OnRequest` part doesn't seem to come into play at all, as in, we don't seem to be testing anything in the `CrashLogging` class. Am I missing something?

AutomatticTracks/src/test/java/com/automattic/android/tracks/CrashLogging/CrashLoggingTest.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    @Test
+    fun `should call for required arguments during initialisation`() {
```
oguzkocer:
```suggestion
    fun `should call for required arguments during initialization`() {
```

Just so it's consistent with the `initialize` method name, even though both spelling seems to be correct. Feel free to ignore it if you end up changing the method name for `initialize`.

AutomatticTracks/src/test/java/com/automattic/android/tracks/CrashLogging/CrashLoggingTest.kt
```diff
@@ -0,0 +1,70 @@
...
+import org.junit.runner.RunWith
+import org.mockito.Mockito.never
+import org.mockito.Mockito.spy
+import org.mockito.Mockito.times
+import org.mockito.Mockito.verify
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.RuntimeEnvironment
+
+@RunWith(RobolectricTestRunner::class)
+class CrashLoggingTest {
```
oguzkocer: After reviewing this class and leaving some suggestions about Kotlin usage, which I think can still be useful as a discussion point, I am failing to understand the testing approach here. These tests read to me as if it's trying to test the implementation details, but it doesn't have a way to do that, so it's trying to fake it. Let me explain.

First of all, the tests are extremely fragile. For example, if I were to add lines with just `dataProvider.applicationContext` and `dataProvider.userContext` to the `CrashLogging.setNeedsDataRefresh` method, I'd get 3 test failures. Of course having those lines by themselves would be meaningless because this would be a read-only action.

The issue is the tests are utilizing the `CrashLoggingDataProvider` interface which by design is just used to provide data. Whether that data is read once or 10 times doesn't matter. You can maybe change those to be `at least once`, in which case they'll provide some value, but I am not sure if that'd be enough or whether they'd match what the test names indicate.

I think the tests would be more useful if we had an access to an interface which prompted to do things. Think of names with actions in them, such as `applyUserTracking`. In that case, we could maybe write a test that says `if this happens make sure applyUserTracking is called once`.

In its current state, I feel like the tests are not there to help the implementation, so I think it's worth considering removing the tests entirely or re-working them to provide more value. I hope that makes sense!

AutomatticTracks/src/androidTest/java/com/automattic/android/tracks/CrashLogging/SendEventsToCrashLoggerService.kt
oguzkocer: I am not sure how I feel about this. I can understand the idea, but it feels odd to me to do it like this. Since I can't provide a meaningful counterpoint not to do it, I guess I'll just leave a couple suggestions in case we keep it:

* The `Service` postfix is [a reserved keyword in Android](https://developer.android.com/guide/components/services) and it really doesn't describe what this class does. So, I suggest using a more descriptive without the `Service` in it. 
* We should document how to use this class and explain why it's in the project in the first place.
* `waitForEventToBeSent` doesn't seem to do anything. I understand it's communicating that the event will take time to appear, but I don't think it should actually cause a delayed finish unless we are taking an action after the delay. We should just have the delay communicated in the documentation I mentioned.
* We shouldn't run these tests in CI. We have a quota in Sentry and they really don't provide any value unless we are actively testing the Sentry integration.

---

While typing the suggestions, I realized why I feel uneasy about this. I think this makes a lot more sense as an example app instead of an instrumentation test. We don't verify anything in the tests, so they provide no value as an automated test. We could potentially add a way to check if the Sentry event is there (using their API), but I don't see enough value in that.

Instead, I think we can have a very simple app within this repo that simply adds Tracks library to it and have buttons in the UI that calls these different methods. I can definitely get behind something like that.

oguzkocer (changes requested):
Sorry about the late review @wzieba. I've been AFK for 3 days last week and only got around to being able to review this. I've left some (possibly) extensive comments/suggestions, I hope you'll find them useful.

Both @jkmassel and I will be busy for the next couple weeks and considering there are quite a few things to talk through, and more PRs to come, I've asked @malinajirka to help us with the reviews. I am sure he'll provide even more value than I can at the moment since he has been actively working in Android.

@malinajirka If you can look over my comments and provide your thoughts as well, especially if I am missing something or you disagree with the suggestion, I'd really appreciate it. I'll definitely follow up with the comments as much as I can, but I consider you to be the primary reviewer now. We really appreciate you helping us with these reviews 🙇‍♂️

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLoggingDataProvider.kt
wzieba: I don't know. This was the api that was used before, I don't have enough knowledge to decide on any change here.

If I was about to guess, I'd say that there might be cases when some value stored as `aplicationContext` _might_ be `null` because why not? Information that something _does not exist_ is still valuable information.

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.kt
wzieba: This way we can see that this value has a deliberate "empty string" value, it's not a mistake or typo. I'm ok with naming it an `EMPTY_STRING`

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.kt
wzieba: I'm totally ok with removing it, I didn't understand its purpose in the first place but I assumed that it's useful for some cases I'm not aware of.

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.kt
wzieba: ~Sure, after your comment I can see that it's also a good candidate for `inline` method~
IDE marked the inlining of this method as `insignificant impact`. I think that in the end R8 should inline it in the place of execution so I'll just simplify it as you suggested.

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLoggingDataProvider.kt
wzieba: I'll link the answer to `applicationContext` as I think it's the same case: https://github.com/Automattic/Automattic-Tracks-Android/pull/51#discussion_r599957914

AutomatticTracks/src/test/java/com/automattic/android/tracks/CrashLogging/CrashLoggingTest.kt
wzieba: > because it leads to the possibility of the value being changed 

I'm not sure if I understand you correctly, could you please ~deliberate~ elaborate on that?

JUnit creates a separate class for every executed test so I don't see a possibility that tests will interfere with each other if you meant that.

I'm ok with refactoring this as you propose, it looks good too.

AutomatticTracks/src/test/java/com/automattic/android/tracks/CrashLogging/CrashLoggingTest.kt
wzieba: > we don't seem to be testing anything in the CrashLogging class

I think we do test something, we test that `CrashLogging#crash` method will throw an exception, which is expected behavior. Naming it `OnRequest` was about `request` from client/library consumer in a sense of "when client wants to". 

Eventually, as discussed here https://github.com/Automattic/Automattic-Tracks-Android/pull/51#discussion_r599870154 I'll remove `CrashLogging#crash` so this test won't cause any confusion anymore.

AutomatticTracks/src/androidTest/java/com/automattic/android/tracks/CrashLogging/SendEventsToCrashLoggerService.kt
wzieba: > The Service postfix is a reserved keyword in Android

Sure, I'll rename that to not cause any confusion

> We should document how to use this class and explain why it's in the project in the first place.

What kind of documentation do you think of? More live Javadoc like, or some instructions on P2 or README on Github? What would suit here the best?

> `waitForEventToBeSent` doesn't seem to do anything (...) We should just have the delay communicated in the documentation I mentioned.

`waitForEventToBeSent` does something and it's locking a main thread to not allow the test app to close immediately after test execution. Otherwise, the events would not be sent because test app will close before HTTP request from Sentry will have a chance to be sent. Mentioning this delay in documentation won't change anything.

> We shouldn't run these tests in CI

And we don't do this, do we? Current CI task is `./gradlew --stacktrace testRelease` and it does not have any dependencies on `connectedAndroidTest` which runs instrumented tests on Android. If we would run this test on CI, even in this build it would fail as there's no setup for the emulator to run on CI.

---

>  I think this makes a lot more sense as an example app instead of an instrumentation test.

I get your point that this class is not a test and I don't want to make this a test. 

I'd opt against having sample app for testing Sentry. Sample app causes a lot of additional overhead like keeping dependencies up to date or writing UI and platform code. Also, typical testing flow will require more effort with rebuilding the app and manually pressing UI buttons. I think this mechanism I've proposed is faster to work with and easier to maintain. To test if everything works fine after changes, developers will only need to run methods from one class.

Do those additional arguments convince you to leave this mechanism (I intentionally avoid word `test`)?

AutomatticTracks/src/test/java/com/automattic/android/tracks/CrashLogging/CrashLoggingTest.kt
wzieba: > For example, if I were to add lines with just dataProvider.applicationContext and dataProvider.userContext to the CrashLogging.setNeedsDataRefresh method, I'd get 3 test failures. 

It's good that you'd get test failures because those fields are not meant to be assigned anywhere in `CrashLogging#setNeedsDataRefresh`. I know it's not ideal (are requested but not assigned so eventually behavior won't change) but it at least indicates that something is changed in a way that was initially implemented.

> I think the tests would be more useful if we had access to an interface that prompted to do things

I get your points. In my opinion test in the current state does provide some value. Thinking of a scenario when some changes will be performed in `CrashLogging`, this test will assert that after changes, the fields that were previously requested, are still requested. Requesting a field is likely to equal setting this in Sentry. I know - not ideal.

I'm okay though to abstract `Sentry` static invocations to some class, this way we'll have a possibility to have more robust test but again, I don't think this test in its current form is problematic or doesn't bring a value.

wzieba:
> Sorry about the late review @wzieba. I've been AFK for 3 days last week and only got around to being able to review this. I've left some (possibly) extensive comments/suggestions, I hope you'll find them useful.

Hi @oguzkocer , no problem at all. I treat working on `Tracks` as a kind of "side quest" so it doesn't block me in any way to be productive in other areas.

I find your comments and suggestions very useful, many thanks for them!

> Both @jkmassel and I will be busy for the next couple weeks and considering there are quite a few things to talk through, and more PRs to come, I've asked @malinajirka to help us with the reviews. I am sure he'll provide even more value than I can at the moment since he has been actively working in Android.

Sure thing, happy to welcome @malinajirka here 👋

AutomatticTracks/src/androidTest/java/com/automattic/android/tracks/CrashLogging/SendEventsToCrashLoggerService.kt
malinajirka: > What kind of documentation do you think of? More live Javadoc like, or some instructions on P2 or README on Github? What would suit here the best?

I personally think a regular comment (javadoc) should be enough.

> waitForEventToBeSent does something and it's locking a main thread to not allow the test app to close immediately after test execution. Otherwise, the events would not be sent because test app will close before HTTP request from Sentry will have a chance to be sent. Mentioning this delay in documentation won't change anything.

Thanks for the clarification! Could you please elaborate on this part "Mentioning this delay in documentation won't change anything.". I personally feel like explaining in a comment why the delay is needed might help other devs/future-us. Wdyt?

> I get your point that this class is not a test and I don't want to make this a test.

Could you please elaborate a bit more on this? I feel like every class which is annotated with "RunWith" is considered as a test by Gradle. For example, if you run "./gradlew cAT" it'll run even methods annotated with @Test in this class, right? I completely agree with Oguz that keeping this as a test might bite us in the future. Having said that, I understand your point of view, that having this "tool" might come handy. 
We might want to consider adding "@Ignore" annotation and a comment explaining that this test is disabled so it's not run automatically because it's a manual test that needs the developer to manually check Sentry that everything works as expected. It's far from optimal, I know. Wdyt?

> And we don't do this, do we? Current CI task is ./gradlew --stacktrace testRelease and it does not have any dependencies on connectedAndroidTest which runs instrumented tests on Android. If we would run this test on CI, even in this build it would fail as there's no setup for the emulator to run on CI.

This is the current state, but we might want to start running connected tests on a CI in the future. I'm worried no-one will realize they need to manually suppress this test from running. Wdyt? 

--------------------

When it comes to the discussion "app" vs "test". I personally don't mind either as long as it's documented - javadoc or readme(for sample app). I think the additional overhead with setting up and maintaining the app will be very low. For me one nice advantage of the app is that anyone who opens the project can see it. As oppose to the test - IMO very few people go through the test suite to find if it contains something interesting. On the other hand, I agree that nice thing about the test is that it runs automatically and is quick - you don't need to manually click on buttons.

You both made some good points and no decision is final and we can always iterate on it if we find that it doesn't work for us.

AutomatticTracks/src/androidTest/java/com/automattic/android/tracks/CrashLogging/SendEventsToCrashLoggerService.kt
```diff
@@ -0,0 +1,54 @@
+package com.automattic.android.tracks.crashlogging
```
malinajirka: I'm still getting oriented in the code, so this might be a stupid question - when I run the "service/test" it fails with `InvalidDsnException`. Could we perhaps add a dummy value into gradle.settings-example and use BuildConfig.xyz to hint the devs they need to use a valid sentry dsn?

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.kt
malinajirka: I agree that removing this seems like a good call.

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLoggingDataProvider.kt
malinajirka: This seems like a minor issue and changing it is probably not worth the risk - considering we call use these fields even from java code, it might not be easy to determine if we ever pass null or not. Wdyt?

AutomatticTracks/src/test/java/com/automattic/android/tracks/CrashLogging/CrashLoggingTest.kt
malinajirka: I see your point that the tests bring some value. At the same time I think the value is very low and I feel like these tests won't help us with refactoring and extending this class. As Oguz mentioned even a small unrelated change to the class breaks 3 tests. 

We are checking if certain fields were read but nothing else - eg. if I remove "Sentry.setUser(user)" the tests will still pass even though the code doesn't work.

IMO unit tests should be trying to verify that certain actions result in the expected result. I'm worried the current approach might make extending this class more difficult instead of facilitating it.  I think the main issue is that the class is not testable at all - tbh the Sentry class is here to blame, since it makes testing way more difficult than it should. All in all, I'd consider removing these tests or trying to make the class testable (wrapping Sentry as you suggested) and writing a new unit test suite.

 > but it at least indicates that something is changed in a way that was initially implemented.

This might end up leading to a completely different discussion, so please feel free to ignore this part completely. But IMO this is something unit tests should never try to do. As mentioned above, IMO unit test should verify that certain action results in the expected result, but should not try to verify the internal implementation. In might experience doing so can do more harm than good.

malinajirka (commented):
Really great job @wzieba and very thorough review and great comments @oguzkocer! Thank you both 🙇 

>  I am sure he'll provide even more value than I can at the moment since he has been actively working in Android.

Unfortunately, reading your comments @oguzkocer I'm already sure I won't provide such a great reviews. But I'll do my best ;).


The code is already in a great state. I shared my two cents on some of the ongoing discussions. Let me know what you think ;).

AutomatticTracks/src/androidTest/java/com/automattic/android/tracks/CrashLogging/SendEventsToCrashLoggerService.kt
wzieba: > Could you please elaborate on this part "Mentioning this delay in documentation won't change anything."

What I meant by writing this is that this method is not about "communicating that the event will take time to appear". It's a vital part of sending a message as tests run on the test app (I mean: created for tests) and for every test execution we've got a new run of the app. After each test is finished app is closed so without locking the main thread (=pausing test for a few seconds) Sentry won't have time to be sent.

Rephrasing this statement to be more descriptive: "Knowing about delay won't resolve any issue here. It's not about informing developer about something - this pause is a must". So this is what I meant: just knowing that there's a delay won't happen that this will work.

> every class which is annotated with "RunWith" is considered as a test by Gradle

You're right, by saying "this class is not a test" I meant that it's not a test in a formal way, not that this is not a test by Gradle standards. 

> We might want to consider adding "@ignore" annotation and a comment explaining that this test is disabled

That's a good idea, I'm starting to drifting into a sample app solution though (more on that in the latter part of this comment) so it might not be needed.

---

Thank you for your input on that ""service"" idea. I can see that this idea raises a lot of concerns so eventually, it might not be a good fit.

Although I'm not in love with the idea of the sample app (I just have aversion to buttons :) ) I'll prepare one to avoid any confusion now and in the future.

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLogging.kt
wzieba: I've decided to suggest the`.orEmpty()` method, I think it satisfies both concerns: there's no constant and there's the explicitness of intention.

AutomatticTracks/src/androidTest/java/com/automattic/android/tracks/CrashLogging/SendEventsToCrashLoggerService.kt
wzieba: Sure, I've extracted it to `gradle.properties-example` in https://github.com/Automattic/Automattic-Tracks-Android/pull/51/commits/056de39c9de7bc0b68152fd0387f35d122f864bc and added documentation in https://github.com/Automattic/Automattic-Tracks-Android/pull/51/commits/9eda370145632b8e1bba55994da6dd8131c08555

AutomatticTracks/src/test/java/com/automattic/android/tracks/CrashLogging/CrashLoggingTest.kt
wzieba: I've majorly refactored this test to make it more robust and to check more things. I like it much better now, thank you for your feedback! You can check it here: https://github.com/Automattic/Automattic-Tracks-Android/pull/51/commits/713aeb434628356839e201ef3ef19ad828e7a987 but later I added so improvements so you may prefer to check the whole branch changes.

> IMO unit test should verify that certain action results in the expected result, but should not try to verify the internal implementation

I think we are both on the same page with that. I see I didn't present my thought as I wanted to: by saying "something is changed" I didn't mean an internal implementation but rather "something changed in behavior". Again - this wording was poorly on my side so sorry for the confusion.

AutomatticTracks/build.gradle
```diff
@@ -24,19 +24,29 @@ configurations {
...
+    testImplementation 'junit:junit:4.13.2'
+    testImplementation "org.assertj:assertj-core:3.19.0"
+
+    androidTestImplementation 'androidx.test.ext:junit:1.1.2'
+    androidTestImplementation 'androidx.test:runner:1.3.0'
 }
 
 apply from: 'ktlint.gradle'
 
+def CONFIG = { property -> "${project.properties.get(property)}" }
```
wzieba: This will lazily evaluate a property from `gradle.properties`

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLogging.kt
```diff
@@ -0,0 +1,101 @@
...
+
+    @JvmStatic
+    fun start(
+        context: Context,
+        dataProvider: CrashLoggingDataProvider,
+    ) {
+        start(context, dataProvider, SentryErrorTrackerProxyImpl())
+    }
+
+    internal fun start(
```
wzieba, line 26: This method allows us to test any logic for `CrashLogging`. I see no other option, as the static nature of `CrashLogging` won't allow us to inject an object `SentryErrorTrackerProxy` in any other way.

I know that production code shouldn't be bothered by tests. In this case though, I believe there are more advantages than harm: this method is `internal` (won't leak to library consumers) and tests written for this class assert, that all fields have been correctly assigned.

Do you have any opinion or suggestion on that?

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLogging.kt
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    @JvmStatic
+    fun log(throwable: Throwable, data: Map<String, String?>) {
+        val event = SentryEvent().apply {
+            message = Message().apply {
+                message = throwable.message
+            }
+            level = SentryLevel.ERROR
+            setExtras(data.toMutableMap() as Map<String, String?>)
```
wzieba, line 92: This map **has** to be a mutable one, otherwise, we'll have a crash in runtime as Sentry SDK tries to edit an immutable map (which is a default in `Kotlin`).

I've reported an issue on Sentry's Github, more details and reproduction repository is there if you're interested: https://github.com/getsentry/sentry-java/issues/1357

AutomatticTracks/src/androidTest/java/com/automattic/android/tracks/CrashLogging/SendEventsToCrashLoggerService.kt
wzieba: If it's okay for you, I've created an issue #69 and I will work on creating a sample app on a separate pull request to not shadow changes regarding Sentry bump.

I think it's okay to leave two options: automated sending of events using Android's instrumentation tests framework and a sample app.

> We might want to consider adding "@ignore" annotation and a comment explaining that this test is disabled

After reconsidering I'd like to suggest not adding `@Ignore` to this "test". A scenario when it'll run alongside "real instrumentation tests" is not a case for now (and I don't suspect it'll change soon) and if it will happen - adding `@Ignore` is a little effort. Is that okay or you'd prefer to add `@Ignore` now?

I use this method quite often and it'd be a bummer to have to add and remove `@Ignore` on every interaction with git.

AutomatticTracks/build.gradle
wzieba: I've removed that, apparently it's not needed.

AutomatticTracks/src/androidTest/java/com/automattic/android/tracks/CrashLogging/SendEventsToCrashLoggerService.kt
malinajirka: > Rephrasing this statement to be more descriptive: "Knowing about delay won't resolve any issue here. It's not about informing developer about something - this pause is a must". So this is what I meant: just knowing that there's a delay won't happen that this will work.

I should have read the comments more carefully 😞 . Thanks for the explanation!

> I think it's okay to leave two options: automated sending of events using Android's instrumentation tests framework and a sample app.
> After reconsidering I'd like to suggest not adding @Ignore to this "test". A scenario when it'll run alongside "real instrumentation tests" is not a case for now (and I don't suspect it'll change soon) and if it will happen - adding @Ignore is a little effort. Is that okay or you'd prefer to add @Ignore now?
I use this method quite often and it'd be a bummer to have to add and remove @Ignore on every interaction with git.

I'm just a bit worried that whoever starts running tests on a CI probably won't even know that they should prevent these specific tests from running. I'm also worried about the precedens - if we use it here, it'll be much easier to use the same approach again in other repos. 
The fact that it's a bit cumbersome is imo the price for using a tool which is not meant to be used for "tests" which require a human interaction. However, I'm wondering if we could perhaps have some setting in gradle.properties instead of using @Ignore. Wdyt @wzieba @oguzkocer ?

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLogging.kt
malinajirka: I wish CrashLogging wasn't an object - static classes/methods just make testing more difficult. I understand that refactoring it into an injectable class is above the scope of this PR. So with the current constraints, I think this approach is ok. 
One thing we might want to consider is using @VisibleForTesting annotation.  It documents the fact that the methods signature was modified for tests and the tools will complain if we use the method from production code. Wdyt?

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryErrorTrackerProxy.kt
```diff
@@ -0,0 +1,23 @@
+package com.automattic.android.tracks.crashlogging.internal
+
+import android.content.Context
+import io.sentry.SentryEvent
+import io.sentry.SentryOptions
+import io.sentry.protocol.User
+
+interface SentryErrorTrackerProxy {
```
malinajirka: I'm wondering if we really need to introduce this interface. Can't we simply work with `SentryErrorTrackerProxyImpl`? 

1. The interface is dependant on Sentry anyway, so it won't help us if we ever decide to replace Sentry with another library.
2. Mockito allows us to mock even classes so this interface imo won't bring any value when it comes to testing.
3. Kotlin compiler is smart, so it won't re-compile files which don't need to be re-compiled (as oppose to Java compiler which afaik always recompiles all affected dependencies).

All in all I'm wondering if we are not just introducing an unnecessary boilerplate. Wdyt?

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryErrorTrackerProxyImpl.kt
```diff
@@ -0,0 +1,41 @@
+package com.automattic.android.tracks.crashlogging.internal
+
+import android.content.Context
+import io.sentry.Sentry
+import io.sentry.SentryEvent
+import io.sentry.SentryOptions
+import io.sentry.android.core.SentryAndroid
+import io.sentry.protocol.User
+
+class SentryErrorTrackerProxyImpl : SentryErrorTrackerProxy {
```
malinajirka: Nitpick 🔍 : I think we usually call classes which just delegate calls to the original object "wrappers" instead of "proxies". Having said that, I agree it's a proxy so I'm fine keeping the name.

AutomatticTracks/src/test/java/com/automattic/android/tracks/crashlogging/CrashLoggingTest.kt
```diff
@@ -1,65 +1,142 @@
...
-
-@RunWith(RobolectricTestRunner::class)
-class CrashLoggingTest {
 
-    lateinit var dataProvider: CrashLoggingDataProvider
-    var user: TracksUser? = null
+class CrashLoggingTest {
 
-    private fun initialize(currentUser: TracksUser? = null) {
+    val fakeProxy = FakeSentryErrorTrackerProxy()
```
malinajirka: ❓ What are advantages of using a fake object instead of a mock in this test suite? If we were using a mock, we could get rid of some boilerplate. I also believe that it'd be more consistent with other tests in our codebases. Wdyt?

AutomatticTracks/src/test/java/com/automattic/android/tracks/crashlogging/CrashLoggingTest.kt
```diff
@@ -0,0 +1,161 @@
...
+import io.sentry.SentryLevel
+import org.assertj.core.api.Assertions.assertThat
+import org.assertj.core.api.SoftAssertions
+import org.junit.Test
+import java.util.Locale
+
+class CrashLoggingTest {
+
+    val fakeProxy = FakeSentryErrorTrackerProxy()
+    val mockedContext = Activity()
```
malinajirka: nitpick: This and the above fields can be private - AS is complaining.

malinajirka (commented):
Thanks for the changes @wzieba! The tests look definitely better now ;). I've left some questions but none of them is blocking - I'm just curious about your thoughts.

AutomatticTracks/src/androidTest/java/com/automattic/android/tracks/CrashLogging/SendEventsToCrashLoggerService.kt
wzieba: > The fact that it's a bit cumbersome is imo the price for using a tool which is not meant to be used for "tests" which require a human interaction

I agree with this and previous points - still, I believe it's a very helpful tool to test this library e2e.

I think I've found a compromise (IMO quite neat!) in 8d74cf24682ee0cb9c9c33d79bc9af8f126d3cde . Current behavior:
- Running `./gradlew connectedAndroidTest` results with `No tests found`
- Running `SendEventsToSentry` from Android Studio will run this test as we explicitly ask to run this class

WDYT? Does it work for you this way?

malinajirka: TIL about `'notClass'` argument! LGTM ;)

wzieba: > TIL about 'notClass' argument!

Me too! From this article if you're interested diving into more details: https://medium.com/stepstone-tech/exploring-androidjunitrunner-filtering-options-df26d30b4f60

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLogging.kt
wzieba: > I wish CrashLogging wasn't an object - static classes/methods just make testing more difficult.

Yes, that's true. I think migrating this to an object isn't difficult for a library, rather for Tracks consumers as we would have to start injecting this object in many places - I don't know if WPAndroid or WCAndroid or SimpleNote are ready for that... 😅 


> One thing we might want to consider is using @VisibleForTesting annotation.

Sure, added in 1f600e65cf2b8f779a812b17cf32f5c3b50c8c1a

malinajirka: > Yes, that's true. I think migrating this to an object isn't difficult for a library, rather for Tracks consumers as we would have to start injecting this object in many places - I don't know if WPAndroid or WCAndroid or SimpleNote are ready for that... 😅

I don't know about WCAndroid and SimpleNote, but it should be a rather simple change for WPAndroid. I think other repos could create a static wrapper if they preferred the static access. So I don't think it'd be too much effort. Having said that it's not focus of this PR so lets not worry about it.

> Sure, added in 1f600e6

Thanks!

wzieba: > I don't know about WCAndroid and SimpleNote, but it should be a rather simple change for WPAndroid

That's sounds good! I'll keep that in mind with future work on Tracks

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryErrorTrackerProxyImpl.kt
wzieba: If it can help anyone later understand this class, I'm cool with changing it to widely used `Wrapper` naming. Updated in [`9f6f595` (#51)](https://github.com/Automattic/Automattic-Tracks-Android/pull/51/commits/9f6f595f619410a9509a20ef17f3fe70a46989aa)

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryErrorTrackerProxy.kt
wzieba: It was introduced to make testing easier (creating a fake object). As I've followed your suggestion in https://github.com/Automattic/Automattic-Tracks-Android/pull/51#discussion_r603352435, I've refactored tests to use mocks so eventually I've removed this interface. 

Answering your questions:
1. Totally right, this interface didn't abstract idea of `ErrorTracker` in general
2. If we use `Mockito` - that's true. For `Fakes`, we wouldn't be able to write test
3. Are you referring to `compile avoidance`? This was not a reason for writing this interface, as Kotlin doesn't support it. Editing `SentryErrorTrackerProxy**Impl**` would cause recompilation of `CrashLogging` in Kotlin (wouldn't if it would be Java). Or do you refer to the incremental compilation? If so, I'm not sure if I understand how it works: maybe the internal implementation of `SentryErrorTrackerProxy**Impl**` changes won't cause recompilation of `CrashLogging`?

AutomatticTracks/src/test/java/com/automattic/android/tracks/crashlogging/CrashLoggingTest.kt
wzieba: I like fakes for several reasons:
1. We don't have to have an additional dependency on `Mockito` and `Mockito inline`
2. Slightly faster test invocation (doesn't matter in such a small repo)
3. It's easier to adjust to any SUT changes (as there's an interface we have to satisfy, not mocks that have to be applied)

The cons is that I had to add a dummy interface we discussed above.

In the case of this repo, I don't have a favor in any direction. I moved to mocks in https://github.com/Automattic/Automattic-Tracks-Android/pull/51/commits/87027acd5715ed1cef55d3ff4a093e16fe6c0c9d as I like we could remove `SentryErrorTrackerProxy` interface.

Thanks for pointing out!

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLogging.kt
```diff
@@ -0,0 +1,101 @@
...
+    ) {
+        this.sentryWrapper = sentryWrapper
+        this.dataProvider = dataProvider
+
+        sentryWrapper.initialize(context) { options ->
+            options.apply {
+                dsn = dataProvider.sentryDSN
+                environment = dataProvider.buildType
+                release = dataProvider.releaseName
+                setDebug(dataProvider.enableCrashLoggingLogs)
```
wzieba, line 39: Setting `BuildConfig.DEBUG`, which I changed from in 58e2187e3a10cebf66665c8c3ad7d0b19fb77a5e, did not make sense as I forgot that this is not project-included library, so `BuildType` of project won't match `BuildType` of library. We release `release` variant on Jitpack so we would always had `setDebug(false)` in clients.

I've updated `CrashLoggingDataProvider` to provide wether client wants to see logs or not.

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryErrorTrackerProxy.kt
malinajirka: > Are you referring to compile avoidance? This was not a reason for writing this interface, as Kotlin doesn't support it. Editing SentryErrorTrackerProxy**Impl** would cause recompilation of CrashLogging in Kotlin (wouldn't if it would be Java). Or do you refer to the incremental compilation? If so, I'm not sure if I understand how it works: maybe the internal implementation of SentryErrorTrackerProxy**Impl** changes won't cause recompilation of CrashLogging?

I don't have a deep understanding of compilers. I just remember a talk where they compared java vs kotlin compiler. One of the points was that introducing interfaces in java to speed up incremental compilation can help in Java but is not necessary in Kotlin, since kotlin compiler is smart enough.

> This was not a reason for writing this interface, as Kotlin doesn't support it.

Hmm, I didn't know that Kotlin doesn't support it. I've read [this article](https://blog.jetbrains.com/kotlin/2020/09/the-dark-secrets-of-fast-compilation-for-kotlin/) and it seems that it supports it. Or do you think it's different for java+kotlin projects? I'm just curious, feel free to ignore this comment if you don't know from the top of your head.

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryErrorTrackerProxyImpl.kt
malinajirka: Thanks!

AutomatticTracks/src/test/java/com/automattic/android/tracks/crashlogging/CrashLoggingTest.kt
malinajirka: Thanks for elaborating on this! 

I agree with your comments, but since we already depend on Mockito and it brings more features, I think moving to mocks was a good call. Thanks!

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLogging.kt
malinajirka: Ohh, great catch! 👏

malinajirka:
Thanks for all the improvements @wzieba!

I think this PR is almost ready to be merged. One last thing I noticed is that the `notClass` argument for JUnit takes affect even when I run the tests from Android Studio. It seems that AS runs the following command
```
adb shell am instrument -w -m -e notClass com.automattic.android.tracks.crashlogging.SendEventsToSentry   
-e debug false -e class 'com.automattic.android.tracks.crashlogging.SendEventsToSentry'
 com.automattic.android.tracks.test/androidx.test.runner.AndroidJUnitRunner
```
which results in a crash since I'm trying to run SendEventsToSentry tests but at the same time telling it that these tests should be ignored. I'm on Android Studio 4.1.2. Can you think of a reason why it doesn't behave this way with your setup?

wzieba:
Thank you @malinajirka for this check! As we discussed on DMs, I had to have an old run configuration when I was checking how it works for me. Good point with noticing that, it could cause confusion for future users. In https://github.com/Automattic/Automattic-Tracks-Android/pull/51/commits/aa536f459baccdffff4784fdf703ec48f36b614c I:
- Added run configuration with `Include Extra Params from Gradle build file` unchecked
- Added small "troubleshooting" part into comment in `SendEventsToSentry` about making sure to uncheck `Include Extra Params from Gradle build file`

AutomatticTracks/src/main/java/com/automattic/android/tracks/CrashLogging/CrashLoggingDataProvider.kt
wzieba: I've ended changing the signature of `applicationContext` and `userContext` to `Map<String, String>`. The reason for that is that `Sentry SDK` won't accept a `Map` with a null value for a map's entry. More details and fix: https://github.com/getsentry/sentry-java/issues/1366

We could leave this interface and map `null` to an empty string or maybe drop all entries that have null value but I'd like to leave this decision to the library consumer.

wzieba:
For the second commit since your last review (https://github.com/Automattic/Automattic-Tracks-Android/pull/51/commits/6a4998a9e28306ed832c97b32436448259bfc60c) @malinajirka I've added a new comment in this thread: https://github.com/Automattic/Automattic-Tracks-Android/pull/51#discussion_r604349015

malinajirka (approved):
All works as expected now. Thanks @wzieba !

PR #55, opened 2021-02-26
Add CI configuration file

### LINKED ISSUE #52: Continuous integration setup
Maybe w could add continuous integration setup to this library, WDYT @oguzkocer @jkmassel ?

I was thinking about two simple workflows/pipelines:
1. Basic check for commit - build and test
2. Release on new tag

If it's ok, I can help with that, I believe CircleCI is our choice for every CI/CD right?

wzieba:
I've verified and run this configuration with [CircleCI CLI](https://circleci.com/docs/2.0/local-cli/).

Resolves: #52


wzieba:
I'm amending and force pushing this commit _without any changes_ to see if it make this branch visible on CircleCI.

Reference: https://discuss.circleci.com/t/no-other-branches-but-develop/19194

Edit: it worked

.circleci/config.yml
```diff
@@ -0,0 +1,25 @@
...
+      api-version: "30"
+    steps:
+      - checkout
+      - android/restore-gradle-cache
+      - run:
+          command: ./gradlew --stacktrace assembleRelease
+      - run:
+          command: ./gradlew --stacktrace testRelease
+      - run:
+          command: ./gradlew --stacktrace publishToMavenLocal
```
wzieba: Sorry for change after requesting review. I've just added this step as checking of publishing to maven local works fine is a useful check in context of migrating to `Jitpack` #56

oguzkocer: I am not sure how I feel about this to be honest. It feels a bit out of context to me and a little unnecessary. I think the `assemble` step should be enough and when/if we start using Jitpack for this repo, we should get warned of any errors. We don't have this step for other projects that uses Jitpack and we didn't have any issues in them this step would help with. Finally, we might not even use Jitpack, it's just a probability.

What do you think @loremattei @jkmassel?

oguzkocer (commented):
@wzieba Thanks for working on this. I left a reply to your comment about `publishToMavenLocal`. Aside from that, I _think_ we'd want to have the lint step, although that'd require us fixing a few lint issues we currently have. What do you think?

.circleci/config.yml
wzieba: > we should get warned of any errors

I don't know how `Jitpack` warns about problems but running `publishToMavenLocal` on CI would prevent us from merging PR that would break on `Jitpack` before any client could use it.

E.g. we've got this scenario at this moment. `publishToMavenLocal` does not work (reason is described in #56 ) but both `assemble` and `test` tasks works fine.

I'm ok with removing this check, I thought it might be a good idea to have extra safety on that and I don't see any negative outcomes.

wzieba:
Thanks for review,

> I think we'd want to have the lint step, although that'd require us fixing a few lint issues we currently have

Sure, that's a good idea. Currently we've got 18 lint errors - I'd opt to define baseline files for them (in scope of this PR) and take care of resolving them later. Does it work for you?

oguzkocer:
> Sure, that's a good idea. Currently we've got 18 lint errors - I'd opt to define baseline files for them (in scope of this PR) and take care of resolving them later. Does it work for you?

Yeap, that sounds good to me. Thank you!

.circleci/config.yml
oguzkocer: After spending some time trying to figure out why Jitpack is being weird, I can see the value in having this task here. Let's give it a shot.

Can we add a comment explaining this is for testing Jitpack integration?

wzieba: I've added names to steps instead of writing comment in 69f04d2a62930370ae6c8ebf6074fe8d0e04c8f8. This way it's visible already in CircleCI logs that this step is about asserting Jitpack build correctness. Does it work for you?

wzieba:
The build will not pass now because of reasons described in #56

.circleci/config.yml
oguzkocer: This looks great, thank you!

oguzkocer (approved):
Thanks @wzieba!

Issue #56, opened 2021-02-26
Prepare project to work with Jitpack


wzieba:
[As we plan to host Tracks on Jitpack](https://github.com/Automattic/Automattic-Tracks-Android/issues/52#issuecomment-785967559), we have to make sure that it'll be compatibile.

Bumping project to use `Gradle 6.1.1` in #54 results with failing `./gradlew publishToMavenLocal` as this task is provided currently by `bintray-release` plugin which [does not support `Gradle 6.x` ](https://github.com/novoda/bintray-release/issues/329)

Failing `./gradlew publishToMavenLocal` results with failing Jitpack build as [this is the way Jitpack prepares artifacts
](https://github.com/jitpack/jitpack.io/blob/master/ANDROID.md#checks).

PR #57, opened 2021-02-26
Migrate to maven-publish. Drop bintray-release

wzieba:
Reasons for this migration are described in ticket #56


AutomatticTracks/build.gradle
```diff
@@ -38,16 +38,24 @@ android {
...
-    dryRun = 'false'
-    autoPublish = 'true'
+afterEvaluate {
+    publishing {
+        publications {
+            maven(MavenPublication) {
+                artifactId 'Automattic-Tracks-Android'
+                artifact("$buildDir/outputs/aar/AutomatticTracks-debug.aar")
+                description 'Android client for Nosara tracks (event tracking and analytics)'
+                groupId 'com.github.Automattic'
```
wzieba: Changing `artifactId` and `groupId` to match convention forced by `Jitpack` so using local maven will require only changing version, not the whole dependency address.

Local maven artifactory might not be that useful after migration to composite builds mentioned [here](https://github.com/Automattic/Automattic-Tracks-Android/pull/51/files#r582846727) but for now I think it adds a convenient way to work with `Tracks` locally.

AutomatticTracks/build.gradle
```diff
@@ -38,16 +38,24 @@ android {
...
+                pom {
+                    licenses {
+                        license {
+                            name = 'GNU General Public License v2.0'
```
wzieba, line 50: This license change aligns to the one defined in project's initial commit https://github.com/Automattic/Automattic-Tracks-Android/commit/5652b539769f6874d81f112fe742a1ec7ad87b6e

wzieba:
Jitpack's build log for the most recent commit: c435538ecb48b96e887827bd4e7107359b20d5fe

https://jitpack.io/com/github/Automattic/Automattic-Tracks-Android/issue~55-use-maven-publish-c435538ecb-1/build.log

wzieba:
and just a heads up - build is failing only because #55 is not yet merged and I enabled CircleCI which starts build on every new push.

oguzkocer:
@wzieba I am a little surprised about this PR. I wasn't actually expecting it to happen right now. More importantly though, I am not sure if most of this is necessary. Have you maybe tried to include the library to Jitpack as is?

If it does work as is, maybe we shouldn't have the `maven-publish` plugin and all the extra information that goes with it. I don't think Jitpack would actually use this information, but I am not 100% sure.

We also probably don't want to mention the `local maven` bit in the README. That's just an implementation detail for Jitpack (and quite a hack) and it shouldn't matter to us. If we can help it, I'd rather not point fingers `mavenLocal` as it's not a recommended tool.

wzieba:
> Have you maybe tried to include the library to Jitpack as is?

Yes I did and I came across problem described in #56 . [Log from failed build on Jitpack edit: on current `develop`](https://jitpack.io/com/github/Automattic/Automattic-Tracks-Android/develop-4fd591e7c8-1/build.log)

If we don't want to use `maven-publish`, we could use some [fork of bintray-release](https://github.com/panpf/bintray-publish) with support of `Gradle 6.x` but as (1) this fork seems unpopular (2) we want to migrate from `Bintray` soon, I thought it might be a good idea to just add support for Jitpack already as we'll need to release new version of `Tracks` soon.

> If we can help it, I'd rather not point fingers mavenLocal as it's not a recommended tool.

I'm ok with removing all additional fields that were used before when `Bintray` was used and I'm ok with removing all bits about local maven from README - I just thought that it might be useful as composite builds are not yet supported with `Tracks`.

oguzkocer:
> Yes I did and I came across problem described in #56 . Log from failed build on Jitpack edit: on current develop

I missed that, sorry. Let me go back and review both PRs with this information.

AutomatticTracks/build.gradle
```diff
@@ -38,16 +38,26 @@ android {
...
-    licences = ['MIT']
-    website = 'https://github.com/Automattic/Automattic-Tracks-Android'
-    dryRun = 'false'
-    autoPublish = 'true'
+afterEvaluate {
+    publishing {
+        publications {
+            maven(MavenPublication) {
+                artifactId 'Automattic-Tracks-Android'
+                artifact("$buildDir/outputs/aar/AutomatticTracks-debug.aar")
```
oguzkocer: I believe we can use `from components.release` instead.
```suggestion
                from components.release
```

AutomatticTracks/build.gradle
oguzkocer: I don't think we should be using local maven. I'd rather just make it possible to use composite builds with this library instead as that should be very straightforward for this library. I think we should just stick to `com.automattic` for the `groupId`.

AutomatticTracks/build.gradle
```diff
@@ -38,16 +38,26 @@ android {
...
+                    licenses {
+                        license {
+                            name = 'GNU General Public License v2.0'
+                            url = 'https://www.gnu.org/licenses/old-licenses/gpl-2.0.en.html'
+                        }
+                    }
+                }
+                if (project.hasProperty("localDistribution") && project.getProperty("localDistribution").toBoolean()) {
+                    version = "LOCAL-SNAPSHOT"
+                }
```
oguzkocer: I think we should remove everything related to local maven from the changeset. I'd be happy to assist with composite builds next week, so we don't need it.

README.md
```diff
@@ -17,23 +17,34 @@ projects but the idea is to share what we've made.
...
-* Publish to bintray:
+* Publish to Jitpack (for `Tracks` >`1.2.1`)
 
-```sh
-$ ./gradlew assemble publishToMavenLocal bintrayUpload -PbintrayUser=XXX -PbintrayKey=XXX -PdryRun=false
-```
-
-Note: running the `publishToMavenLocal` task is very important, it will
-create the .po file needed for any maven repository.
+There's no need to do anything. Jitpack will handle build of requested verison itself.
```
oguzkocer: We probably don't need this section anymore. I don't think it adds any value.

README.md
```diff
@@ -17,23 +17,34 @@ projects but the idea is to share what we've made.
...
 
 ## Usage
 
+### Jitpack
```
oguzkocer: It might be worth adding a link to Jitpack build instructions in this section in case our README gets out of sync. (not necessarily to this line)

README.md
```diff
@@ -17,23 +17,34 @@ projects but the idea is to share what we've made.
...
+}
+```
+
+### Maven Local
+
+Run `./gradlew publishToMavenLocal -PlocalDistribution=true` and change version to `LOCAL-SNAPSHOT` on client:
+
+```gradle
+dependencies {
+    implementation 'com.github.Automattic:Automattic-Tracks-Android:LOCAL-SNAPSHOT'
```
oguzkocer: Similar to my previous comment, I think we should remove this section as well.

oguzkocer (changes requested):
I really thought we didn't need the `maven-publish` plugin as we don't use it in repositories such as FluxC and Jitpack works just fine. However, I couldn't figure out why that's the case in the limited time I've spent on it. (still more than I really should have)

Anyway, I left some suggestions, most of which is about mavenLocal. I just don't think it's a good idea to get that into the mix and I'd personally much rather spend a couple hours setting up the composite build support for this project. Having said that, if you want to keep all of it for now, I am OK with that.

Finally, I think it'd be a good idea to merge `origin/develop` into this branch just to verify that everything works with Gradle 6. (for full disclosure, I already tried that in a personal branch to verify the `from components.release` change, so it's not strictly necessary)

wzieba:
> I really thought we didn't need the maven-publish plugin as we don't use it in repositories such as FluxC and Jitpack works just fine. However, I couldn't figure out why that's the case in the limited time I've spent on it. (still more than I really should have)

Thank you for spending time on this and other PRs. I've done some research and come back to you with answers. Sorry if some of those points are redundant to you, I wanted to be extra clear.

### Why `FluxC` works without `publishToMavenLocal` task defined (so without `maven-publish`)?

#### How Jitpack builds

Jitpack, before building will look for building tasks. The first choice is `publishToMavenLocal` but the second one is `install` and this is the task that is used in Jitpack to build `FluxC`.
See `Getting tasks: ./gradlew tasks --all` in Jitpack logs of [`Tracks` with `publishToMavenLocal` task](https://jitpack.io/com/github/Automattic/Automattic-Tracks-Android/issue~55-use-maven-publish-c435538ecb-1/build.log) and [`FluxC` with `install` task](https://jitpack.io/com/github/wordpress-mobile/WordPress-FluxC-Android/1.12.0-beta-2/build.log)). 

If Jitpack won't find any task, it'll attempt to add android and maven plugins and try to run `install` task ([see this logs](https://jitpack.io/com/github/Automattic/Automattic-Tracks-Android/test~without_bintray_plugin-0569b00835-1/build.log)).

#### Why `FluxC` has `install` task defined?
`FluxC` uses plugin `android-maven-gradle-plugin` [which adds `install` task](https://github.com/dcendents/android-maven-gradle-plugin/blob/develop/src/main/java/org/gradle/api/plugins/AndroidMavenPlugin.java#L239~L244) that, copying from task description:
> Installs the 'archives' artifacts into the local Maven repository.

🎯 

##### What happens if we remove `android-maven-gradle-plugin` ?

Only Gradle modules that are not android libraries will be installed (`processor` and `annotations`) and this is because they both have `maven` plugin applied. Jitpack build log [from `FluxC` without `android-maven-gradle-plugin`](https://jitpack.io/com/github/wzieba/WordPress-FluxC-Android/test~remove_android_maven_gradle-1.12.0-beta-2-gd1db226-10/build.log) and from [`FluxC` without both `android-maven-gradle-plugin` *and* `maven` ](https://jitpack.io/com/github/wzieba/WordPress-FluxC-Android/test~remove_amg_and_maven_plugins-1.12.0-beta-2-g925d4b4-11/build.log). The second one has artificially added `maven` plugin (I mean: added by Jitpack during build) and only not android-based modules are built.

### Proposals

After this research I can suggest a few possible ways to handle publishing `Tracks` on Jitpack.

#### Proposal 1: Use `maven-publish`

So basically the solution proposed in this PR with taking into account your comment reviews. I like this idea because it follows [Jitpack's guide](https://jitpack.io/docs/ANDROID/#gradle):
> To enable building on JitPack you need to configure the maven-publish Gradle plugin to publish your library as explained in the Android documentation.

#### Proposal 2: Use `android-maven-gradle-plugin` 

This would benefit with the same setup as is in `FluxC` but [this plugin is abandoned](https://github.com/dcendents/android-maven-gradle-plugin) with last code-related change 3 years ago. The authors suggest to use `maven-publish`.

To preview this proposal please check a57195454880668468e38be39496e75d0af1eeac and [log from `Jitpack` ](https://jitpack.io/com/github/Automattic/Automattic-Tracks-Android/test~android_maven_plugin-a571954548-1/build.log)

#### ~Proposal 3: Use our own build task~

I was thinking about running `assemble` task to generate .aar artifact and then copy it to `~/.m2/repository` but Jitpack requires `pom.xml` too and in order to generate it we would have to apply some plugin like `maven-publish` so I think there's not much sense. [Jitpack log from this experiment](https://jitpack.io/com/github/Automattic/Automattic-Tracks-Android/test~own_jitpack_task-d6a24bf500-1/build.log) and commit d6a24bf500d4ec67b0a279819102867313fa0226

----

I believe the proposal 1 is the best in this scenario but I understand reason you might want to choose proposal 2 (same setup across different projects). Please ping me if you decide on preferable solution and I'll adjust to that. Thanks!

wzieba:
One more "proposal", just FYI as it doesn't work.

#### ~Proposal 4: Do not add any maven publish related plugins~

We could remove `bintray-release` and `maven` plugins and let Jitpack add needed plugins on its own but I don't suggest it as (1) it seems dangerous in terms of build stability to let 3rd party modify source code before running (2), it simply does not work (at least out of the box). [Log from failed build](https://jitpack.io/com/github/Automattic/Automattic-Tracks-Android/test~no_publish_plugins-07ed3ac7f1-1/build.log) and commit with change 07ed3ac7f1f85f418b1892cfe0978d8c88132f44

README.md
wzieba: That's a good idea. I've ended up with removing custom instruction I wrote and adding link to Jitpack instruction as duplicating this indeed seems redundant. Does it work for you?

oguzkocer (approved):
Thanks for all the research @wzieba! I was looking for the `install` task in FluxC, but missed that plugin in my haste. I agree that the first solution looks to be the best in this case and if we switch away from Jitpack, we might even utilize it ourselves. Sorry it took some time for me to wrap my head around the differences!

Note that, although CI is failing right now, we decided to address it after merging these PRs.

PR #65, opened 2021-03-16
Set up ktlint

### LINKED ISSUE #63: Integrate static code analysis for Kotlin
We could enable static code analysis for Kotlin, a good candidate could be ktlint which [is used in other libs, e.g. FluxC ](https://github.com/wordpress-mobile/WordPress-FluxC-Android#using-ktlint)

wzieba:
Resolves #63 

Basic setup for [ktlint](https://github.com/pinterest/ktlint) in sync with [WooCommerce](https://github.com/woocommerce/woocommerce-android/blob/develop/build.gradle#L50-L66) or [WordPress](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/build.gradle#L117-L133) codebases.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,6 +14,9 @@ jobs:
           name: Setup gradle.properties
           command: cp gradle.properties-example gradle.properties
       - run:
+          name: Static code analysis
+          command: ./gradlew --stacktrace ktlint
+      - run:
           name: Assemble .aar
           command: ./gradlew --stacktrace assembleRelease
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ jobs:
           command: cp gradle.properties-example gradle.properties
       - run:
           name: Static code analysis
-          command: ./gradlew --stacktrace ktlint
+          command: ./gradlew --stacktrace ciktlint
       - run:
           name: Assemble .aar
           command: ./gradlew --stacktrace assembleRelease
--- a/AutomatticTracks/build.gradle
+++ b/AutomatticTracks/build.gradle
@@ -18,10 +18,17 @@ repositories {
     jcenter()
 }
 
+configurations {
+    ktlint
+}
+
 dependencies {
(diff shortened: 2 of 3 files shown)


oguzkocer (approved):
@wzieba Thanks for setting this up. As a next step, it'd be great to split these up into separate jobs, such as `Build`, `Lint`, `Test` etc. If you're up for that, you can probably use the WCAndroid config file as a reference. If not, we can do that at a later time :)

wzieba:
@oguzkocer thanks for reviewing. 

>  it'd be great to split these up into separate jobs, such as Build, Lint, Test

Sure, I can do that but what's the reason behind it? I can see in [WCAndroid](https://github.com/woocommerce/woocommerce-android/blob/develop/.circleci/config.yml) for every job we have to perform checkout, `gradle.properties` setup, cache-in and cache-out.

Do you know if those caches are per-job or per-workflow? If per-job, then we lose Gradle's incremental builds (with up-to-date tasks) and Gradle's build cache. If per-workflow, we don't but we still use resources to pass cache between jobs.

What are the benefits of keeping those tasks (`Build`, `Lint`, `Test`) in separate CircleCI jobs?

oguzkocer:
> What are the benefits of keeping those tasks (Build, Lint, Test) in separate CircleCI jobs?

The main benefit for me would be to have individual flows failing, so we can target what needs to be fixed. The secondary benefit is the jobs will run in parallel.

> Do you know if those caches are per-job or per-workflow? If per-job, then we lose Gradle's incremental builds (with up-to-date tasks) and Gradle's build cache. If per-workflow, we don't but we still use resources to pass cache between jobs.

The caches should be per-workflow if I understand this correctly: https://circleci.com/docs/2.0/caching/

---

My suggestion comes more from "let's setup this the right way from the beginning" and less so from the performance gains. Having said that this is probably going to stay as a small repo, so maybe it's not worth it.

@jkmassel What do you think? I don't have as much experience with CircleCI setups, so I'd love to get some input from another MIE.

PR #66, opened 2021-03-16
Resolve lint errors/warnings

### LINKED ISSUE #58: Resolve Android lint errors
In #55 (precisely: dbc594f3008fc61a07591695e385f226d4b45dfb) `lint-baseline` file has been introduced, which suppress several Android lint errors. We should resolve them and remove the baseline file.

wzieba:
Resolves #58 

This PR resolves most of Android lint errors and warnings. I'm leaving two of them:
- `Recycle` in `SqlUtils.java:100` #68 
- `StaticFieldLeak` in `TracksDatabaseHelper.java:25` #67 

as those are not trivial without some more refactoring.


AutomatticTracks/src/main/java/com/automattic/android/tracks/MessageBuilder.java
```diff
@@ -29,7 +30,7 @@
     public static final String ALIAS_USER_ANONID_PROP_NAME = "anonId";
 
     public static synchronized boolean isReservedKeyword(String keyToTest) {
-        String keyToTestLowercase = keyToTest.toLowerCase();
+        String keyToTestLowercase = keyToTest.toLowerCase(Locale.ROOT);
```
wzieba, line 33: Docs on `Locale.ROOT`:
> Useful constant for the root locale.  The root locale is the locale whose
    language, country, and variant are empty ("") strings.  This is regarded
    as the base locale of all locales, and is used as the language/country
    neutral locale for the locale sensitive operations.

In this context is a good choice as those data-related operations are not related to any real `Locale`.

AutomatticTracks/build.gradle
```diff
@@ -39,6 +41,7 @@ android {
 
     lintOptions {
         baseline file("lint-baseline.xml")
+        warningsAsErrors true
```
wzieba, line 49: This helps in lint warning awareness. Lint warnings are super easy to overlook (there's usually just one line on CI that looks like this: `Ran lint on variant release: 2 issues found`). 

Marking warnings as errors shouldn't slower a development process as there's baseline file defined so it is still possible to pass a CI build with warnings.

AutomatticTracks/build.gradle
```diff
@@ -4,6 +4,7 @@ buildscript {
         jcenter()
     }
     dependencies {
+        //noinspection GradleDependency
```
oguzkocer: Can we make `GradleDependency` issues warnings instead? I think you can do something like this:

```
lintOptions {
        baseline file("lint-baseline.xml")
        warningsAsErrors true
        warning 'GradleDependency'
}
```


or better yet, we can do these in a `lint.xml` file.

AutomatticTracks/src/main/java/com/automattic/android/tracks/DeviceInformation.java
```diff
@@ -389,14 +389,12 @@ public Boolean isBluetoothEnabled() {
...
         String bluetoothVersion = null;
-        if (Build.VERSION.SDK_INT >= 8) {
-            bluetoothVersion = "none";
-            if(Build.VERSION.SDK_INT >= 18 &&
-                    mContext.getPackageManager().hasSystemFeature(PackageManager.FEATURE_BLUETOOTH_LE)) {
-                bluetoothVersion = "ble";
-            } else if(mContext.getPackageManager().hasSystemFeature(PackageManager.FEATURE_BLUETOOTH)) {
-                bluetoothVersion = "classic";
-            }
+        bluetoothVersion = "none";
```
oguzkocer: Nitpick: We can combine these 2 lines.

oguzkocer (changes requested):
@wzieba Thanks for taking care of these. I've left a couple minor comments.

AutomatticTracks/build.gradle
wzieba: I'm not sure if I understand - `GradleDependency` [is a lint warning](http://tools.android.com/tips/lint-checks). I just suspended it here as we intentionally don't want to bump Android Gradle Plugin (to have it in sync with other projects as you explained) and I suspended `sentry` bump as there's separate PR for that.

I could disable this warning at all but I believe it is a useful one. What do you think?

wzieba: sorry, I think just understood. You want to override severity once again. So it's warning -> error (`warningAsErrors`) -> warning again (because of `warning`). Cool idea, didn't know it.

I can do that, sure 👍

wzieba: I wasn't able to do this in `lint.xml` file. It looks like this file is interpreted before `lintOptions` extension is configured. This results in having `GradleDependency` as a warning in the moment of evaluating `warningsAsErrors` so eventually it'll be an error.

Maybe I'm missing something, could you make it work with `lint.xml`?

oguzkocer: I haven't tried it with `lint.xml` either, but I think the order might need to be reversed. In any case, I don't think that should block the PR as the current lint configuration is not that extensive.

oguzkocer (approved):

PR #70, opened 2021-03-31
Add sample app

### LINKED ISSUE #69: Prepare sample app
A sample app will help us in end-to-end, manual testing of the library.

wzieba:
Resolve #69 

### Description

This PR adds a new module: test app for helping in end-to-end tests integration of `Automattic-Tracks-Android` with `Sentry`

### Demo
https://user-images.githubusercontent.com/5845095/113126016-82207380-9217-11eb-8666-f22dcc9a8cd2.mov


settings.gradle
```diff
@@ -1 +1,20 @@
-include ':AutomatticTracks'
+pluginManagement {
```
wzieba, line 1: [Plugins DSL](https://docs.gradle.org/current/userguide/plugins.html#sec:plugins_block) is the favored way of declaring Gradle plugins, in opposite to `apply plugin:''`

Here I declare versions of `Android Gradle Plugin` and `Kotlin Android` plugin for both modules: `lib` and `sampletracksapp`

settings.gradle
```diff
@@ -1 +1,20 @@
...
+        google()
+        gradlePluginPortal()
+        jcenter()
+    }
+    resolutionStrategy {
+        eachPlugin {
+            if (requested.id.id.contains("com.android")) {
+                useModule("com.android.tools.build:gradle:4.0.2")
+            }
+        }
```
wzieba, line 15: It's required as `AGP` doesn't support [Plugin Markers Artifacts](https://docs.gradle.org/current/userguide/plugins.html#sec:plugin_markers). [It'll be fixed in AGP 4.2.0 ](https://issuetracker.google.com/issues/64551265#comment10)

AutomatticTracks/lib/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryErrorTrackerWrapper.kt
```diff
@@ -0,0 +1,41 @@
+package com.automattic.android.tracks.crashlogging.internal
+
+import android.content.Context
+import io.sentry.Sentry
+import io.sentry.SentryEvent
+import io.sentry.SentryOptions
+import io.sentry.android.core.SentryAndroid
+import io.sentry.protocol.User
+
+internal class SentryErrorTrackerWrapper {
```
wzieba: This file is not changed - it's just moved (I'm not sure why git treats this file differently)

malinajirka:
Thanks for working on this @wzieba! It looks great overall and the app works as expected. I just have a couple questions:
1. I'm curious why you decided to create a "lib" folder under "AutomatticTracks" instead of creating the example app in a completely separate module. Are there any benefits to this setup? I'm asking mostly because AFAIK we don't use this approach in other repos.
2. "SentryErrorTrackerWrapper.kt" is placed out of its module - "AutomatticTracks/src/main/..." instead of "AutomatticTracks/lib/src/main/...". I think that's the reason why it appears as a new file on GitHub.

wzieba:
> 1. I'm curious why you decided to create a "lib" folder under "AutomatticTracks" instead of creating the example app in a completely separate module.

No apparent reason and I can't see any benefits of that. I'll flatten this in the next commit. Thanks!

> 2. "SentryErrorTrackerWrapper.kt" is placed out of its module

Right! Now I see that I even duplicated this file. Thanks, the next commit removes it.

malinajirka:
Thanks for the changes @wzieba  ;)

> No apparent reason and I can't see any benefits of that. I'll flatten this in the next commit. Thanks!

We might want to consider keeping the library itself in the main `AutomatticTracks` module. 
1. This PR wouldn't make that many changes, it'd simply contain changes related to introduction of the sample app module
2. Anyone who clones the repo, could clearly see what is the core of this repo
3. I feel like "lib" folder/module is usually used for external dependencies, but it's not the case for this repo. It currently contains the core.

I'm sorry for not presenting all my thoughts during the first review 😞 . Wdyt?

wzieba:
> We might want to consider keeping the library itself in the main AutomatticTracks module.

Sure, I'm okay with any naming convention. I'm not especially happy with `AutomatticTracks` name (we know it's Automattic from `groupId`). This PR is not about changing a name though, I'm okay with leaving this name as it is.

> This PR wouldn't make that many changes,

This argument especially appeal to me 👍

malinajirka (approved):
LGTM, thanks! 🚢

PR #72, opened 2021-04-01
Update interface to reflect mutable data cases

### LINKED ISSUE #71: Additional logging data are immutable
## Issue

Interface introduced in #51 has been changed from function invocations to fields. Having a sample from `WCAndroid` implmenetaion, applying field like:

```kotlin
override val currentUser: TracksUser = TracksUser(
        currentAccount?.userId.toString(),
        currentAccount?.email,
        currentAccount?.userName
    )
```

will result with no updating a user. The `currentUser` will be evaluated only once. Client's could resolve this by using Kotlin's backing fields like:

```kotlin
override val currentUser: TracksUser
      get() = TracksUser(
          currentAccount?.userId.toString(),
          currentAccount?.email,
          currentAccount?.userName
      )
```

but this is highly unpleasant API to work with and for sure error-prone.

## Enhancement

`Tracks` offers a method `CrashLogging#setNeedsDataRefresh`. So in addition to updating data, clients need to signal (again) an intention to update it. I'd like to suggest that responsibility to apply updated data should be on `Tracks` side.

Instead of 
```kotlin
fun setCurrentSite(site: SiteModel?) {
      this.currentSite = site
      CrashLogging.setNeedsDataRefresh()
  }
```

I'd like clients to work like this

```kotlin
fun setCurrentSite(site: SiteModel?) {
      this.currentSite = site
  }
```

wzieba:
Resolves: #71 

## Motivation

Described in #71 

### Update user case
1. Make sure `sentryTestProjectDSN` is set to some test project on Sentry in `gradle.properties`.
2. Run `SendEventsToSentry#logMessageWithUpdatedUser`

#### Expected result
There's one new `NullPointerException` event with 2 users. 

One of them has user data with `1` prefix, the second event has user data with `2` prefix.

image

### Update application context case
1. Make sure `sentryTestProjectDSN` is set to some test project on Sentry in `gradle.properties`.
2. Run `SendEventsToSentry#logMessageWithAppendedApplicationContext`

#### Expected result

There are 3 new `OutOfMemory` events. Looking at those events `ADDITIONAL DATA` section, there should be 3 cases. Event with:

1. `1 application : context`
2. `1 application : context` and `2 application : context`
3. `1 application : updated context` and `2 application : context`


AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLogging.kt
```diff
@@ -39,41 +47,25 @@ object CrashLogging {
...
                 setTag("locale", dataProvider.locale?.language ?: "unknown")
                 beforeSend = SentryOptions.BeforeSendCallback { event, _ ->
-                    return@BeforeSendCallback if (dataProvider.userHasOptedOut) {
+                    return@BeforeSendCallback if (userOptOut) {
                         null
                     } else {
-                        event
+                        event.apply {
+                            user = dataProvider.userProvider()?.toSentryUser()
+                        }
```
wzieba, line 51: This is inspired by the current implementation on `WPAndroid`. It removes the responsibility from the client to update the state of current user data. Instead, before every event, a correct user is applied.

https://github.com/wordpress-mobile/WordPress-Android/blob/a9eeed79f6d85fd3a3a55157ff9b2932248c00cc/WordPress/src/main/java/org/wordpress/android/util/CrashLogging.kt#L36-L42

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLogging.kt
```diff
@@ -39,41 +47,25 @@ object CrashLogging {
...
                     }
                 }
             }
         }
-        setNeedsDataRefresh()
-    }
-
-    @JvmStatic
-    fun setNeedsDataRefresh() {
```
wzieba, line 54: This is one of the key points of this PR. I'd prefer to model `Tracks` API in a way that client is not responsible to force-refresh data. It should just provide data or update them if somethings changed.

AutomatticTracks/src/main/java/com/automattic/android/tracks/TracksUser.kt
```diff
@@ -0,0 +1,8 @@
+package com.automattic.android.tracks
+
+data class TracksUser(
+    val userID: String,
+    val email: String,
+    val username: String,
+    val context: Map<String, String>,
```
wzieba: I've decided to move `userContext` into `TracksUser` but I'm thinking about removing it at all. 

- [`WPAndroid` doesn't use it](https://github.com/wordpress-mobile/WordPress-Android/blob/a9eeed79f6d85fd3a3a55157ff9b2932248c00cc/WordPress/src/main/java/org/wordpress/android/util/CrashLogging.kt#L23)
- [`WCAndroid` deosn't use it](https://github.com/woocommerce/woocommerce-android/blob/9e3ee2b151a19d3f2ddf9979147ebaa17fbffee4/WooCommerce/src/main/kotlin/com/woocommerce/android/util/CrashUtils.kt#L68)
- [`simplenote-android` doesn't use it](https://github.com/Automattic/simplenote-android/blob/25c4eccdd4f153ad6a5dda7b4bb7d8de427f337b/Simplenote/src/main/java/com/automattic/simplenote/utils/CrashUtils.java#L94)

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryErrorTrackerWrapper.kt
```diff
@@ -15,14 +14,6 @@ internal class SentryErrorTrackerWrapper {
         }
     }
 
-    fun clearBreadcrumbs() {
-        Sentry.clearBreadcrumbs()
-    }
-
-    fun setUser(user: User?) {
-        Sentry.setUser(user)
-    }
```
wzieba, line 24: Friendly reminder: the user is now set in `BeforeSendCallback` https://github.com/Automattic/Automattic-Tracks-Android/pull/72/files#diff-92307566b0dd043337c1aa764dfc72fcb338e57770bd7ffae0a6e52a250fa4f9R50

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLogging.kt
```diff
@@ -39,41 +47,25 @@ object CrashLogging {
...
-                User().apply {
-                    email = tracksUser.email
-                    username = tracksUser.username
-                    others = dataProvider.userContext
-                        .plus("userID" to tracksUser.userID)
-                }
-            }
-        )
+    fun updateUserOptOutPreference(userHasOptOut: Boolean) {
+        this.userOptOut = userHasOptOut
```
malinajirka: ❓ I'm wondering if we could use the same "dataProvider" concept even for this field or do you think it's an overkill?

AutomatticTracks/src/main/java/com/automattic/android/tracks/TracksUser.kt
malinajirka: I agree that it seems the filed isn't being used so we might want to consider removing it. 

One question not related to this PR: Why do we use "TracksUser" for CrashLogging (Tracks vs Sentry). I find having these two tools within one library quite confusing, but using TracksUser just in Sentry and not in Tracks part of the library is even more confusing. I think we might want to consider renaming it and moving it into the "CrashTracking" part of the library. Wdyt?

AutomatticTracks/src/androidTest/java/com/automattic/android/tracks/crashlogging/SendEventsToSentry.kt
```diff
@@ -23,39 +26,63 @@ import org.junit.runner.RunWith
...
+        CrashLogging.log(OutOfMemoryError())
+
+        CrashLogging.appendApplicationContext(mapOf("1 application" to "updated context"))
+        CrashLogging.log(OutOfMemoryError())
+    }
+
+    companion object {
+        @AfterClass
+        @JvmStatic
+        fun tearDown() {
```
malinajirka, line 78: Nice improvement ;)

malinajirka (commented):
Thanks @wzieba! Great improvements! I've left a few questions, let me know what you think.

AutomatticTracks/src/main/java/com/automattic/android/tracks/TracksUser.kt
wzieba: > I find having these two tools within one library quite confusing 

I second to that 👍 

> using TracksUser just in Sentry and not in Tracks part of the library is even more confusing

and to that. Good observation, I'll rename it and move to `crashlogging` package. Thank you!

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLogging.kt
wzieba: At first, I was thinking about the usage of the API. Clients would probably prefer to statically change the flag when a user opts out. 

But after rethinking I can see that this falls into the same trap as setting `TracksUser` - we don't want clients to bother with such updates. Go idea with moving this to `dataProvider` and request it when needed, I'll do this. Thanks!

wzieba:
Sorry for the last commit's message, TIL git doesn't like a **`** character in commit message

malinajirka (approved):
LGTM, thanks for the changes!

Issue #75, opened 2021-04-06
`Tracks` API should be in sync with current `WPAndroid` requirements


wzieba:
`WPAndroid` has its own `Sentry` configuration. `Automattic-Tracks` should cover all needed cases

- [x] filters out the last exceptions that occurred during an `EventBus` event. This should be extracted to `Tracks` library: https://github.com/wordpress-mobile/WordPress-Android/blob/a9eeed79f6d85fd3a3a55157ff9b2932248c00cc/WordPress/src/main/java/org/wordpress/android/util/CrashLogging.kt#L44-L55
---
- [x] allows to toggle `session tracking` feature https://github.com/wordpress-mobile/WordPress-Android/blob/a9eeed79f6d85fd3a3a55157ff9b2932248c00cc/WordPress/src/main/java/org/wordpress/android/util/CrashLogging.kt#L31

**Update**: `session tracking` is enabled by default: https://github.com/getsentry/sentry-java/blob/7209e3700159ca5e52c1825b7d99004edf10e8a6/sentry/src/main/java/io/sentry/SentryOptions.java#L193

---

- [x] allows to define cache dir path https://github.com/wordpress-mobile/WordPress-Android/blob/a9eeed79f6d85fd3a3a55157ff9b2932248c00cc/WordPress/src/main/java/org/wordpress/android/util/CrashLogging.kt#L30

**Update**: `cache dir path` is already set by Sentry SDK. See: https://github.com/getsentry/sentry-java/blob/7209e3700159ca5e52c1825b7d99004edf10e8a6/sentry-android-core/src/main/java/io/sentry/android/core/AndroidOptionsInitializer.java#L208-L212

---

- [x] allows to append data before sending an event https://github.com/wordpress-mobile/WordPress-Android/blob/a9eeed79f6d85fd3a3a55157ff9b2932248c00cc/WordPress/src/main/java/org/wordpress/android/util/CrashLogging.kt#L68-L79

---

- [x] Make `CrashLogging` not a static class (`object`)

---

- [x] API for adding `breadcrumbs` https://github.com/wordpress-mobile/WordPress-Android/blob/a9eeed79f6d85fd3a3a55157ff9b2932248c00cc/WordPress/src/main/java/org/wordpress/android/util/CrashLogging.kt#L96-L101

---

[...]

PR #77, opened 2021-04-06
Drop exception from event's stack trace if requested

wzieba:
This PR adds a feature of dropping the last exception if requested.

For a reason for this feature see `WPAndroid` implementation: https://github.com/wordpress-mobile/WordPress-Android/blob/a9eeed79f6d85fd3a3a55157ff9b2932248c00cc/WordPress/src/main/java/org/wordpress/android/util/CrashLogging.kt#L46-L48

### How to test

I don't have any test cases, the unit test should cover added feature + it's almost the same as current working implementation on `WPAndroid`

### Other ways?

We could offer a `Sentry API` to the clients by using `api` Gradle configuration like:

```groovy
    //implementation 'io.sentry:sentry-android:4.3.0'
    api 'io.sentry:sentry-android:4.3.0'
```

so the client would have an access to `SentryException` and apply the kind of logic presented in this PR on its own but I feel it would cause more harm. @malinajirka if you have other opinions on this, we can discuss it further in the comments.

References: #75


AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLoggingDataProvider.kt
```diff
@@ -28,6 +28,12 @@ interface CrashLoggingDataProvider {
      */
     val enableCrashLoggingLogs: Boolean
 
+    /**
+     * Provides [CrashLogging] with information about exception that should be dropped if is the
+     * last one on stacktrace
+     */
+    val toDropIfLastException: CrashLoggingException?
```
malinajirka: 1. 🔍 Nitpick: I think it'd be good to add an example into the comment why the client app might want to do that. I think the EventBus might be a good example, wdyt?

2.❓Question: Have you considered using a function instead of a field here? Something like:
`shouldDropWrappingException(module: String, type: String, value: String): Boolean`

I think it might be a bit more extensible since the client could drop as many different wrappers as they wanted. Wdyt?

malinajirka (commented):
Thanks @wzieba! I've left a suggestion for you to consider. Let me know what you think ;).

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLoggingDataProvider.kt
wzieba: > the client could drop as many different wrappers as they wanted

Totally agree, I'll change to a function. Thanks!

AutomatticTracks/src/sharedTest/kotlin/com/automattic/android/tracks/fakes/FakeDataProvider.kt
```diff
@@ -12,9 +11,9 @@ class FakeDataProvider(
     override val releaseName: String = "testReleaseName",
     override val locale: Locale? = Locale.US,
     override val enableCrashLoggingLogs: Boolean = true,
-    override val toDropIfLastException: CrashLoggingException? = null,
     var user: CrashLoggingUser? = testUser1,
     var userHasOptOut: Boolean = false,
+    var shouldDropException: (String, String, String) -> Boolean = { _: String, _: String, _: String -> false },
```
wzieba, line 16: `(String, String, String)` doesn't look very pretty but it's used only internally. I can suggest either `typealias` or `inline class` or leaving it as it is. WDYT @malinajirka ?

malinajirka: I think it's fine for a FakeDataProvider. We might want to consider replacing the fake provider with a mocked one in the future, but it's not a priority and perhaps not even a good idea.

malinajirka (approved):
LGTM, thanks!

PR #79, opened 2021-04-07
Drop `JCenter` repository

### LINKED ISSUE #78: Drop `jcenter()` artifactory
Since Jetbrains moved `org.jetbrains.trove4j:trove4j` to `mavenCentral` today, we'll able to drop `jcenter()` artifactory from the project.

https://youtrack.jetbrains.com/issue/IDEA-261387

wzieba:
Resolves #78

### DIFF
--- a/AutomatticTracks/build.gradle
+++ b/AutomatticTracks/build.gradle
@@ -6,7 +6,7 @@ plugins {
 
 repositories {
     google()
-    jcenter()
+    mavenCentral()
 }
 
 dependencies {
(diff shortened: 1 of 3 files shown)


AutomatticTracks/build.gradle
```diff
@@ -6,7 +6,7 @@ plugins {
 
 repositories {
     google()
-    jcenter()
+    mavenCentral()
```
wzieba, line 9: This is already specified in project's `build.gradle` ([here](https://github.com/Automattic/Automattic-Tracks-Android/blob/develop/build.gradle#L4)) so it would work without this line. 

I added `mavenCentral()` explicitly as `Sentry` is published in this repository. I wanted to provide all needed repositories for a module.

malinajirka (approved):
LGTM

PR #80, opened 2021-04-07
Allow clients to append data to an event before sending it

wzieba:
References #75 

This PR adds a feature to append data to every event before sending it.

It'll be primarily used for the `Encrypted Logging` feature but I've extended it to pass a `List<String>` so it doesn't limit to only one use case.

See https://github.com/wordpress-mobile/WordPress-Android/blob/a9eeed79f6d85fd3a3a55157ff9b2932248c00cc/WordPress/src/main/java/org/wordpress/android/util/CrashLogging.kt#L68-L79

### How to test?

1. Make sure there's a `sentryTestProjectDSN` in `gradle.properties` pointing to some playground Sentry project.
2. Run `SendEventsToSentry#shouldAppendExtraForKeyIfRequested`
3. Assert there's a new event with "key |  value" under `additional data` 
image


AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLoggingDataProvider.kt
```diff
@@ -50,4 +50,16 @@ interface CrashLoggingDataProvider {
      * of crash logging data collection
      */
     fun userHasOptOutProvider(): Boolean
+
+    /**
```
malinajirka, line 54: 💡 Idea: Have you considered providing a single method, something like `fun updateEventExtrasBeforeSend(extras: Map<String, Object>): Map<String, Object>`? The library would provide a map of key-value pairs and it'd be up to the consumer to set whatever they want. It feels a bit weird that the consumer needs to inform the library about possible keys first and then the library iterates over these keys and asks the client what values to set. Wdyt?

AutomatticTracks/src/test/java/com/automattic/android/tracks/crashlogging/CrashLoggingTest.kt
```diff
@@ -218,6 +223,44 @@ class CrashLoggingTest {
...
+            appendBeforeSendAction = { updatedExtraValue }
+        )
+
+        val updatedEvent = beforeSendModifiedEvent(capturedOptions, testEventWithExtraApplied)
+
+        assertThat(updatedEvent?.getExtra(extraKey)).isEqualTo(extraValue)
+    }
+
+    @Test
+    fun `should not modify event if it's not meant to be sent`() {
```
malinajirka: 🔍  Nitpick: The "not meant to be sent" part might be a bit confusing for newcomers. What if we renamed this to "should not modify events if the user has opted out"?

malinajirka (commented):
Thanks @wzieba! The changes look good to me and work as described. I've left some suggestions, let me know what you think.
P.S. Feel free to merge the PR if you don't implement any of the suggestions.

malinajirka (approved):

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLoggingDataProvider.kt
wzieba: I get your point @malinajirka . It's a good suggestion unfortunately current Sentry API doesn't allow that.

To make sure if I understood you correctly: by writing 
> The library would provide a map of key-value pairs

you mean `map of key-value pairs` of the `SentryEvent` right? So accessor to it is package-private. See: https://github.com/getsentry/sentry-java/blob/4.3.0/sentry/src/main/java/io/sentry/SentryEvent.java#L264

image

I'll ask Sentry team about that, maybe they will consider making this property a `public` one but it will probably take some time, until the next Sentry release.

malinajirka: I checked the method before writing the comment but I missed the missing "public" modifier 😞 . 

Nitpick: However, I'm still wondering if it isn't cleaner to support `fun getEventExtrasBeforeSend(): Map<String, Object>` and let the consumer return a map and then iterate over the map and set the properties to Sentry. Wdyt?

Feel free to leave it as is and merge this PR, it's approved ;).

wzieba: > I'm still wondering if it isn't cleaner to support fun getEventExtrasBeforeSend(): Map<String, Object>

We have to know current extra values before preparing new one. E.g. `encrypted logs` feature will send encrypted logs and then receive `UUID` for those encrypted logs identification. If we won't check for existing `encrypted logs` `UUID` in `SentryEvent`, we would end up sending those logs more than once (this case is described [here](https://github.com/wordpress-mobile/WordPress-Android/blob/a9eeed79f6d85fd3a3a55157ff9b2932248c00cc/WordPress/src/main/java/org/wordpress/android/util/CrashLogging.kt#L57-L66)).

Your comment inspired me though to pass all `known keys to value` map in signature of a method that prepares extras to `SentryEvent`. This way e.g. we can decide in client if we want to assing a new value for existing key, or if we want to just pass it further. https://github.com/Automattic/Automattic-Tracks-Android/pull/80/commits/878def0cfcf489b35ff15622ba6779018e4e16f2

I'll re-request review as this is quite different from initially proposed. Let me know what you think!

malinajirka (approved):
LGTM, thanks! ;)

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLoggingDataProvider.kt
malinajirka: Thanks for the clarification and improvements ;)!

PR #81, opened 2021-04-07
Convert CrashLogging from object to class

wzieba:
References #75 

This change was discussed in [one of the threads](https://github.com/Automattic/Automattic-Tracks-Android/pull/51#discussion_r603420394) on Github.

What are the advantages of `class` over `object`?

- clients won't have to do [tricks like this one](https://github.com/woocommerce/woocommerce-android/blob/5e8ca2f636f92260d29c5467256df1c4a3def812/WooCommerce/src/main/kotlin/com/woocommerce/android/util/DateUtils.kt#L370) to make classes that use `CrashLogging` testable
- there's no need for the `CrashLogging#start` method. The initialization now happens in the class constructor (`init{}`)
- extracting an interface makes mocking/faking even easier in clients tests - no need to bother with real implementation

Unfortunately, I don't see a way to drop a second constructor with the signature `(Context, CrashLoggingDataProvider, **SentryErrorTrackerWrapper**)` so I left it with `internal` visibility.

### DIFF
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryCrashLogging.kt
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryCrashLogging.kt
@@ -1,15 +1,14 @@
-package com.automattic.android.tracks.crashlogging
+package com.automattic.android.tracks.crashlogging.internal
 
 import android.content.Context
-import androidx.annotation.VisibleForTesting
-import com.automattic.android.tracks.crashlogging.internal.SentryErrorTrackerWrapper
-import com.automattic.android.tracks.crashlogging.internal.toSentryUser
+import com.automattic.android.tracks.crashlogging.CrashLogging
+import com.automattic.android.tracks.crashlogging.CrashLoggingDataProvider
 import io.sentry.SentryEvent
 import io.sentry.SentryLevel
 import io.sentry.SentryOptions
 import io.sentry.protocol.Message
 
-class SentryCrashLogging @VisibleForTesting internal constructor(
+internal class SentryCrashLogging constructor(
     context: Context,
     private val dataProvider: CrashLoggingDataProvider,
     private val sentryWrapper: SentryErrorTrackerWrapper,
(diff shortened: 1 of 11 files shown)


malinajirka:
> Unfortunately, I don't see a way to drop a second constructor with the signature (Context, CrashLoggingDataProvider, **SentryErrorTrackerWrapper**) so I left it with internal visibility.

Could you please elaborate on this a bit more. I'm wondering if we could make `SentryCrashLogging` class internal and provide a static method for creating an instance eg in CrashLogging 
```
companion object {
        fun createInstance(context: Context, dataProvider: CrashLoggingDataProvider) : CrashLogging
         = SentryCrashLogging(context, dataProvider, SentryErrorTrackerWrapper()) 
    }
```
We wouldn't even need two constructors with this approach and the consumer wouldn't even need to know about Sentry. 
 (as a bonus we'd get rid of @VisibleForTesting annotation) Wdyt?

wzieba:
Thanks! This is a great idea, added in https://github.com/Automattic/Automattic-Tracks-Android/pull/81/commits/9b468f7c2b149a24cee2052344dac5ad9f440ec2

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryCrashLogging.kt
```diff
@@ -0,0 +1,76 @@
...
+import io.sentry.SentryOptions
+import io.sentry.protocol.Message
+
+internal class SentryCrashLogging constructor(
+    context: Context,
+    private val dataProvider: CrashLoggingDataProvider,
+    private val sentryWrapper: SentryErrorTrackerWrapper,
+) : CrashLogging {
+
+    constructor(
```
malinajirka, line 17: 🔍 Nitpick: I think we can remove this constructor now.

malinajirka (approved):

malinajirka:
Thanks for the changes @wzieba  ;) Look great! I've left one minor comment, feel free to merge the changes after you remove the unused constructor.

wzieba:
@malinajirka sorry, I forgot to turn off auto-merge. I'll resolve your comment in next PR, hope that's ok!

malinajirka:
@wzieba Interesting, I turned it off 2 seconds after I approved the changes and I thought it wouldn't merge it automatically, but I was wrong. No worries at all - it was my bad actually :).

PR #83, opened 2021-04-09
Change `userHasOptOutProvider` to `crashLoggingEnabled`

### LINKED ISSUE #82: Change `userHasOptOutProvider` to `crashLoggingEnabled`
Relates to: https://github.com/Automattic/Automattic-Tracks-Android/pull/80#discussion_r609058195

#### Reasoning

There are other cases for not wanting to enable crash logging. User opting out is one of them, but there are other factors like type of build. See: https://github.com/wordpress-mobile/WordPress-Android/blob/a9eeed79f6d85fd3a3a55157ff9b2932248c00cc/WordPress/src/main/java/org/wordpress/android/util/CrashLogging.kt#L163-L171

We should update method name to reflect the real behavior and let client decide when to enable/disable crash logging.

wzieba:
Reasoning for this change is described in #82 

Closes: #82


malinajirka (approved):
LGTM

PR #86, opened 2021-04-12
Map `null` values of `SentryException` to empty string

### LINKED ISSUE #85: `java.lang.NullPointerException: lastException.module must not be null`
Sometimes, exceptions that are bundled with `SentryEvent` can have a `null` value for `module`, `type` or `value`.

Those values are not annotated as `@NonNullable` in Sentry SDK. See https://github.com/getsentry/sentry-java/blob/4.3.0/sentry/src/main/java/io/sentry/protocol/SentryException.java#L88

In current implementation we will receive `Null Pointer Exception` and event will be sent without appending any additional data.

wzieba:
Resolves: #85 

There is a second possibility to resolve this: change signature of `CrashLoggingDataProvider#shouldDropWrappingException` from `(String, String, String)` to `(String?, String?, String?)`. I have no opinion about which would be better. I think the proposed one is just fine.

This PR doesn't require any testing, unit test is a sufficient assertion.


malinajirka:
I like this solution more than changing the signature. It all LGTM thanks!

malinajirka (approved):

PR #88, opened 2021-04-13
Use `tags` instead of `extra` for application context

### LINKED ISSUE #87: Use `tags` instead of `extra` for application context
`Sentry`'s `extra` feature is softly detracted. One of its flaws is that it's not searchable. So e.g. when [`WCAndroid` uses it to set `site id`](https://github.com/woocommerce/woocommerce-android/blob/9e3ee2b151a19d3f2ddf9979147ebaa17fbffee4/WooCommerce/src/main/kotlin/com/woocommerce/android/util/CrashUtils.kt#L57-L65), we can't search for all issues from a selected `site`.

Other clients usages:
- `WPAndroid` doesn't use `extra`
- [as well as `simplenote-android`](https://github.com/Automattic/simplenote-android/blob/25c4eccdd4f153ad6a5dda7b4bb7d8de427f337b/Simplenote/src/main/java/com/automattic/simplenote/utils/CrashUtils.java#L88)

Instead, we should use `tags` which should be a perfect use case. 

e.g. this is sample with custom `1-application` tag:
image

wzieba:
Closes: #87 

Reasons for this change are described in the issue.

### How to test
1. Make sure there's `sentryTestProjectDSN` of the test project in `gradle.properties`
2. Run `logMessageWithAppendedApplicationContext`
3. Make sure there are 3 events reported:
- one with `1-application: context` tag
- one with `2-application: context` tag
- one with `1-application: updated context` tag

### Expected test result

image

### DIFF
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLoggingDataProvider.kt
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLoggingDataProvider.kt
@@ -65,6 +65,11 @@ interface CrashLoggingDataProvider {
         currentExtras: Map<ExtraKnownKey, String>,
         eventLevel: EventLevel
     ): Map<ExtraKnownKey, String>
+
+    /**
+     * Provides the {@link CrashLogging} with information about the current application state.
+     */
+    fun applicationContextProvider(): Map<String, String>
 }
 
 typealias ExtraKnownKey = String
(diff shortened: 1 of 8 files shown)


AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLoggingDataProvider.kt
```diff
@@ -65,6 +65,11 @@ interface CrashLoggingDataProvider {
         currentExtras: Map<ExtraKnownKey, String>,
         eventLevel: EventLevel
     ): Map<ExtraKnownKey, String>
+
+    /**
+     * Provides the {@link CrashLogging} with information about the current application state.
+     */
+    fun applicationContextProvider(): Map<String, String>
```
wzieba, line 72: Here I also applied idea with `provider` which requests for data for before sending any event. This is with sync with other providers and results with really nice interface of `CrashLogging`

malinajirka (approved):
Works as described ;)

PR #89, opened 2021-04-14
Support `breadcrumbs`, update interface methods

### LINKED ISSUE #75: `Tracks` API should be in sync with current `WPAndroid` requirements

wzieba:
Resolves #75 

### Context

This PR adds support for `breadcrumbs` information and updates `CrashLogging` interface to make it similar to the one in `WPAndroid`. I think it's much easier to use this way.

I use word `breadcrumbs` in interfaces as it seems it's not a `Sentry`-related word. I don't know if it's common knowledge, I just mention it as TIL 😅 
image

### How to test

This PR introduces a change of interface and a new feature so it's worth to check if every future works correctly. To make it easy to test now and in the future for any regression tests, I've prepared a strategy which makes testing really quick (**~ 5 minutes**). I'm open to any feedback to this idea - was it painless as I intended to make it?

1. Make sure there's `sentryTestProjectDSN` in `gradle.properties`
2. Go to `Sentry` dashboard of test project. Check all of the events, go to menu (3 dots) and delete all issues (to de-clutter the dashboard)
3. Run `SendEventsToSentry` (the whole class test)
4. Go back to `Sentry` dashboard.
5. Select any event. **Read a message that is applied to an event.** Make sure the event fulfills a requirement described in the message
6. Close the issue. Repeat step 5 until there's no events to check

### DIFF
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLogging.kt
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLogging.kt
@@ -1,7 +1,43 @@
 package com.automattic.android.tracks.crashlogging
 
 interface CrashLogging {
-    fun log(throwable: Throwable)
-    fun log(throwable: Throwable, data: Map<String, String?>)
-    fun log(message: String)
+
+    /**
+     * Records a breadcrumb during the app lifecycle but doesn't report an event. This basically
+     * adds more context for the next reports created and sent by [sendReport] or an unhandled
+     * exception report.
+     *
+     * @param[message] The message to attach to a breadcrumb
+     * @param[category] An optional category
+     */
+    fun recordEvent(
+        message: String,
+        category: String? = null
+    )
+
+    /**
+     * Records a breadcrumb with exception during the app lifecycle but doesn't report an event.
+     * This basically adds more context for the next reports created and sent by [sendReport] or an
+     * unhandled exception report.
+     *
+     * @param[exception] The message to attach to a breadcrumb
+     * @param[category] An optional category
+     */
+    fun recordException(
+        exception: Throwable,
+        category: String? = null
+    )
+
+    /**
+     * Sends a new event to crash logging service
+     *
+     * @param[exception] An optional exception to report
+     * @param[tags] Tags attached to event
+     * @param[message] An optional message attached to event
+     */
+    fun sendReport(
+        exception: Throwable? = null,
+        tags: Map<String, String> = emptyMap(),
+        message: String? = null
+    )
 }
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryEventExt.kt
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryEventExt.kt
@@ -1,0 +1,9 @@
+package com.automattic.android.tracks.crashlogging.internal
+
+import io.sentry.SentryEvent
+
+internal fun SentryEvent.appendTags(tags: Map<String, String>) {
+    tags.forEach { (key, value) ->
+        this.setTag(key, value)
+    }
+}
--- a/.run/recordEvent().run.xml
+++ b/.run/recordEvent().run.xml
@@ -1,8 +1,8 @@
 <component name="ProjectRunConfigurationManager">
-  <configuration default="false" name="logMessageWithUpda...()" type="AndroidTestRunConfigurationType" factoryName="Android Instrumented Tests" nameIsGenerated="true">
+  <configuration default="false" name="recordEvent()" type="AndroidTestRunConfigurationType" factoryName="Android Instrumented Tests" nameIsGenerated="true">
     <module name="Automattic-Tracks-Android.AutomatticTracks" />
     <option name="TESTING_TYPE" value="3" />
-    <option name="METHOD_NAME" value="logMessageWithUpdatedUser" />
+    <option name="METHOD_NAME" value="recordEvent" />
     <option name="CLASS_NAME" value="com.automattic.android.tracks.crashlogging.SendEventsToSentry" />
     <option name="PACKAGE_NAME" value="" />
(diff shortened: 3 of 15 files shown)


AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLogging.kt
```diff
@@ -1,7 +1,43 @@
...
+     * adds more context for the next reports created and sent by [sendReport] or an unhandled
+     * exception report.
+     *
+     * @param[message] The message to attach to a breadcrumb
+     * @param[category] An optional category
+     */
+    fun recordEvent(
+        message: String,
+        category: String? = null
+    )
```
wzieba, line 16: `breadcrumbs` for an event looks like this:
image

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLogging.kt
```diff
@@ -1,7 +1,43 @@
...
+     * This basically adds more context for the next reports created and sent by [sendReport] or an
+     * unhandled exception report.
+     *
+     * @param[exception] The message to attach to a breadcrumb
+     * @param[category] An optional category
+     */
+    fun recordException(
+        exception: Throwable,
+        category: String? = null
+    )
```
wzieba, line 29: `breadcrumbs` for an exception looks like this:
image

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLogging.kt
```diff
@@ -1,7 +1,43 @@
 package com.automattic.android.tracks.crashlogging
 
 interface CrashLogging {
-    fun log(throwable: Throwable)
-    fun log(throwable: Throwable, data: Map<String, String?>)
-    fun log(message: String)
```
wzieba, line 6: A new interface for `CrashLogging` is inspired by current `WPAndroid` implementation. [You can check it here.](https://github.com/wordpress-mobile/WordPress-Android/blob/70ce7de8817020d29768074bb408688436320220/WordPress/src/main/java/org/wordpress/android/util/CrashLogging.kt#L87-L161)

An important thing for me when thinking of names was to make it easy for developers to **distinguish** which method _creates and sends_ an event vs. which method _adds something to breadcrumbs_ and doesn't send event **yet**.

I intentionally dropped word `log` as this is difficult to distinguish (to me) if it's about sending an event or not.

A request: @malinajirka could you please check names of new `CrashLogging` methods (and maybe docs above them) and tell if it was easy to understand what do they do and what's difference between them? I'm totally biased at this point so it's hard to tell for me. Thanks!

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryEventExt.kt
```diff
@@ -0,0 +1,9 @@
+package com.automattic.android.tracks.crashlogging.internal
+
+import io.sentry.SentryEvent
+
+internal fun SentryEvent.appendTags(tags: Map<String, String>) {
+    tags.forEach { (key, value) ->
+        this.setTag(key, value)
+    }
+}
```
wzieba, line 9: [`SentryEvent#setTags` will override all previous tags](https://github.com/getsentry/sentry-java/blob/4.3.0/sentry/src/main/java/io/sentry/SentryBaseEvent.java#L139). I've created this extension as we add tags of `application context` as well as we can add tags for a single event.

malinajirka (approved):
Thanks @wzieba! Nice improvements, LGTM 🚢

PR #94, opened 2021-05-05
Fix a bug with `"null"` string instead of `null` value when applying extras to event

### LINKED ISSUE #93: Setting `"null"` String instead of `null` for known key in `SentryEvent`
Current logic applies `"null"` String value for a known key before every event. 

So when we have `knownKeys = listOf("foobar")`, and we won't provide any logic in `DataProvider#provideExtrasForEvent`, we'll end up with `event.get("foobar") = "null"` which is invalid. It should be `event.get("foobar") = null`

wzieba:
Closes: #93

### DIFF
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryCrashLogging.kt
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryCrashLogging.kt
@@ -3,6 +3,7 @@ package com.automattic.android.tracks.crashlogging.internal
 import android.content.Context
 import com.automattic.android.tracks.crashlogging.CrashLogging
 import com.automattic.android.tracks.crashlogging.CrashLoggingDataProvider
+import com.automattic.android.tracks.crashlogging.ExtraKnownKey
 import com.automattic.android.tracks.crashlogging.eventLevel
 import io.sentry.Breadcrumb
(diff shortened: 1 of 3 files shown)


AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryCrashLogging.kt
```diff
@@ -55,8 +56,11 @@ internal class SentryCrashLogging constructor(
         )
     }
 
-    private fun mergeKnownKeysWithValues(event: SentryEvent) = dataProvider.extraKnownKeys()
-        .associateWith { knownKey -> event.getExtra(knownKey).toString() }
```
wzieba, line 59: This is core problem with previous implementation. When `SentryEvent` didn't have applied value for a `knownKey`, we received `null` which we later mapped to string (like `null.toString()`) so we've ended up with `"null"` instead of `null`

wzieba:
Hi @malinajirka 👋  

can I ask you for review of this PR? It's very small and also doesn't require any any tests - unit test checks for the expected behavior and I think it's sufficient check.

malinajirka (approved):
Thanks @wzieba! LGTM ;)

wzieba:
Thank you a lot, I can always count on you! 🙇

PR #97, opened 2021-06-07
Offer `SentryOkHttpInterceptor` to client

### LINKED ISSUE #90: Offer `SentryOkHttpInterceptor` to the client to add automated breadcrumbs for http calls
We can add `SentryOkHttpInterceptor` which will automatically add breadcrumbs with information about all http calls.

https://docs.sentry.io/platforms/android/configuration/integrations/okhttp/

Blocked by https://github.com/wordpress-mobile/WordPress-FluxC-Android/issues/1961

wzieba:
Closes: #90 

This PR adds a feature of offering `SentryOkHttpInterceptor` outside the library. Integrating this interceptor adds some features to crash reporting like automatically added network-related breadcrumbs or performance logs:

### Sample
Screenshot 2021-06-07 at 14 51 13

Because of `sentry-android-okhttp` dependency, I had to set `minSdk` to `21` and this required removing some conditional code.

I think bumping `minSdk` is not a problem as all clients use version `21` or higher: 
- [WPAndroid](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/build.gradle#L121)
- [WCAndroid](https://github.com/woocommerce/woocommerce-android/blob/develop/WooCommerce/build.gradle#L59)
- [simplenote-android](https://github.com/Automattic/simplenote-android/blob/develop/Simplenote/build.gradle#L36)

### DIFF
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLoggingOkHttpInterceptorProvider.kt
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLoggingOkHttpInterceptorProvider.kt
@@ -1,7 +1,8 @@
 package com.automattic.android.tracks.crashlogging
 
 import io.sentry.android.okhttp.SentryOkHttpInterceptor
+import okhttp3.Interceptor
 
 object CrashLoggingOkHttpInterceptorProvider {
-    fun createInstance(): SentryOkHttpInterceptor = SentryOkHttpInterceptor()
+    fun createInstance(): Interceptor = SentryOkHttpInterceptor()
 }
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLoggingOkHttpInterceptorProvider.kt
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLoggingOkHttpInterceptorProvider.kt
@@ -1,0 +1,7 @@
+package com.automattic.android.tracks.crashlogging
+
+import io.sentry.android.okhttp.SentryOkHttpInterceptor
+
+object CrashLoggingOkHttpInterceptorProvider {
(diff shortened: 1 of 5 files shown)


wzieba:
👋 @malinajirka since you've been reviewing most of the previous Tracks PRs, I wonder if you're okay with reviewing this one and a few next ones? Is that okay for you or you would prefer for someone else to contribute to those reviews?

malinajirka:
Sure thing ;)!

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLoggingOkHttpInterceptorProvider.kt
```diff
@@ -0,0 +1,8 @@
+package com.automattic.android.tracks.crashlogging
+
+import io.sentry.android.okhttp.SentryOkHttpInterceptor
+import okhttp3.Interceptor
+
+object CrashLoggingOkHttpInterceptorProvider {
+    fun createInstance(): Interceptor = SentryOkHttpInterceptor()
```
malinajirka, line 7: ❓ This is not used anywhere, so clients won't be able to use this yet, right? I assume this is planned for the upcoming PRs.

malinajirka (approved):
LGTM, thanks ;)

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/CrashLoggingOkHttpInterceptorProvider.kt
wzieba: It's not used anywhere in the lib, but just adding this to the list of interceptors in client will start adding breadcrumbs. Here's snippet of how it'll look on `WCAndroid`: 
image

malinajirka: Ahh, I see. Thanks for the clarification! ;)

PR #105, opened 2021-12-08
Change Sentry version variable to fix `Dependabot`. Bump Sentry to `5.4.3`.

### LINKED ISSUE #104: Fix `Dependabot`
After defining version via Gradle's local property (`def`) in https://github.com/Automattic/Automattic-Tracks-Android/pull/99 , Dependabot stopped creating update PRs, even if in logs it can see that it should.

wzieba:
Closes: #104 

You can see result in fork: https://github.com/wzieba/Automattic-Tracks-Android/pulls

### DIFF
--- a/AutomatticTracks/build.gradle
+++ b/AutomatticTracks/build.gradle
@@ -10,7 +10,7 @@ repositories {
 }
 
 dependencies {
-    implementation platform('io.sentry:sentry-bom:5.0.1')
+    implementation platform('io.sentry:sentry-bom:5.4.3')
     implementation "io.sentry:sentry-android"
     implementation "io.sentry:sentry-android-okhttp"
     implementation 'androidx.annotation:annotation:1.1.0'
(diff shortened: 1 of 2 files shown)


AutomatticTracks/build.gradle
```diff
@@ -10,10 +10,8 @@ repositories {
 }
 
 dependencies {
-    def sentryVersion = "5.0.1"
```
wzieba, line 13: ~There is some issue with `sentry-android-okhttp` that Dependabot can't create PR for it so creating a single variable for two dependencies is problematic.~

I realised Sentry offers BoM. Refactored to it.

malinajirka: I'm wondering if/how BoM works across modules. We sometimes define a version in the project build file and use it in multiple modules. But I guess we could still define bom in the project level build.gradle (feels a bit weird). Anyway, it's fine for this use case since it fixes dependabot, but I'd discuss it with the team before we start using BoM everywhere.

malinajirka (approved):
LGTM, thanks!

malinajirka (approved):
I've tested SendEventsToSentry and SentryCrashLoggingTest and all seemed to be working just fine. Thanks!

PR #114, opened 2022-03-04
Add ExPlat implementation to :experimentation module

### LINKED ISSUE #112: Extract ExPlat from WordPress-Android to Automattic-Tracks-Android
Extract [ExPlat implementation](https://github.com/wordpress-mobile/WordPress-Android/tree/trunk/WordPress/src/main/java/org/wordpress/android/util/experiments) from WordPress-Android to Automattic-Tracks-Android so it can also be used by [woocommerce-android](https://github.com/woocommerce/woocommerce-android) and other projects.

RenanLukas:
Resolves #112

Copies ExPlat implementation from `WordPress-Android` so it can be used by other projects that depend on `Automattic-Tracks-Android`.


experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
```diff
@@ -0,0 +1,89 @@
...
+import org.wordpress.android.util.AppLog.T
+
+class ExPlat(
+    private val experiments: Set<Experiment>,
+    private val experimentStore: ExperimentStore,
+    private val appLogWrapper: AppLogWrapper,
+    private val coroutineScope: CoroutineScope,
+    private val isDebug: Boolean
+) {
+    private val platform = Platform.WORDPRESS_ANDROID
```
wzieba: I believe this should be declared by client

experimentation/src/main/java/com/automattic/android/experimentation/example/ExampleExperiment.kt
```diff
@@ -0,0 +1,7 @@
+package com.automattic.android.experimentation.example
+
+import com.automattic.android.experimentation.ExPlat
+import com.automattic.android.experimentation.Experiment
+
+@Suppress("unused")
+class ExampleExperiment(exPlat: ExPlat) : Experiment(name = "example_experiment", exPlat = exPlat)
```
wzieba: If it's not used anywhere, maybe we can remove it to not cause confusion?

experimentation/src/test/java/com/automattic/android/experimentation/ExPlatTest.kt
```diff
@@ -0,0 +1,243 @@
...
+import org.wordpress.android.fluxc.model.experiments.Variation
+import org.wordpress.android.fluxc.model.experiments.Variation.Control
+import org.wordpress.android.fluxc.model.experiments.Variation.Treatment
+import org.wordpress.android.fluxc.store.ExperimentStore
+import org.wordpress.android.fluxc.store.ExperimentStore.OnAssignmentsFetched
+import org.wordpress.android.fluxc.utils.AppLogWrapper
+import java.util.Date
+
+@ExperimentalCoroutinesApi
+class ExPlatTest {
```
wzieba, line 29: Thanks, love the coverage!

experimentation/build.gradle
```diff
@@ -0,0 +1,55 @@
...
+        versionName "1.0"
+
+        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+    }
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
+    }
+    lintOptions {
+        baseline file("lint-baseline.xml")
```
wzieba: As it's a new module, the baseline is not needed

experimentation/build.gradle
```diff
@@ -0,0 +1,55 @@
...
+android {
+    compileSdkVersion 32
+
+    defaultConfig {
+        minSdkVersion 21
+        targetSdkVersion 30
+        versionCode 1
+        versionName "1.0"
+
+        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
```
wzieba: As this module has no instrumentation tests, let's remove this to not add anything not needed.

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
RenanLukas: Great catch! Thank you

experimentation/src/test/java/com/automattic/android/experimentation/ExPlatTest.kt
RenanLukas: Thank you for implementing it @renanferrari 😁

RenanLukas:
Thank you for the initial review @wzieba ! I've addressed your comments.

Any ideas on why CI is complaining about `:AutomatticTracks`?

wzieba:
@RenanLukas I'm investigating this issue, for this moment:
- lint started to fail on `Cursor#getColumnIndex`
- the quick fix is a change: `Cursor#getColumnIndex` -> `Cursor#getColumnIndexOrThrow`
- this happens in different projects as well (e.g. https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2295/commits/979125fce552767b179f3c01611615a64d21b237)
- the weird thing is that it happens also if you checkout some older version of the library (e.g. commit `a957ae9`, when it was passing few months ago when it was building on CI). The same thing is reported by [different people](https://stackoverflow.com/questions/69053061/android-studio-value-must-be-%E2%89%A5-0#comment122160428_69053061). I'm wondering what is the case, because version of `Android Lint` is associated with `Android Gradle Plugin` so without any code changes it shouldn't break 🤔 . 

Either way this issue is not directly associated with your code changes. I believe it'd be awesome to fix it in a separate PR but ideally it'd be awesome to find a root cause - `EventTable` shouldn't be touched by any means, nothing really changed in implementation.

experimentation/build.gradle
```diff
@@ -0,0 +1,52 @@
...
+        content {
+            includeGroup "org.wordpress"
+            includeGroup "org.wordpress.fluxc"
+            includeGroup "org.wordpress.wellsql"
+        }
+    }
+}
+
+android {
+    compileSdkVersion 32
```
wzieba: @RenanLukas generally, the core issue is that `compileSdkVersion` of this module, influenced which version of lint/lint rules will be used. So lowering it down to `30` will resolve the issue. I think the best will be to bump `compileSdkVersion` of both modules in a seperate PR.

Here is branch on which I was testing, see that build is passing on CI: https://github.com/Automattic/Automattic-Tracks-Android/compare/issue/112-extract-explat...issue/112-testing?expand=1

RenanLukas: Thank you very much @wzieba 👍

renanferrari (approved):
Thanks for handling this! 

LGTM 🚢

wzieba (approved):
Looks good to me!

PR #117, opened 2022-03-12
Adds Buildkite & Publishing to S3

oguzkocer:
This PR adds Buildkite and publishing to S3 following the conventions we have in other projects. Unfortunately both Gradle and Android Gradle Plugin are outdated in this project which meant that I was not able to add publishing sources as part of this PR. Instead of trying to combine that into this PR, I think it'll be better to review & merge this PR as is and add publishing the sources after we have a chance to upgrade Gradle & Android Gradle Plugin.

The CircleCI failure is due to this PR adding `publish-to-s3` Gradle plugin as the `publishToMavenLocal` requires the plugin to be configured. I've already added all the necessary tasks from CircleCI to the new Buildkite configuration, so we can ignore the error and remove CircleCI in a subsequent PR.

I've updated the `README` with the new publishing & usage instructions. I'll open client PRs to update their usage once this PR is merged.

**To Test:**

As long as Buildkite is green, we should be good for this PR. If there are any issues during usage, which is not very likely, I'll follow up with it as a follow up PR.

### DIFF
--- a/AutomatticTracks/build.gradle
+++ b/AutomatticTracks/build.gradle
@@ -1,7 +1,7 @@
 plugins {
     id 'com.android.library'
     id 'org.jetbrains.kotlin.android'
-    id 'maven-publish'
+    id 'com.automattic.android.publish-to-s3'
 }
 
 repositories {
@@ -63,10 +63,14 @@ afterEvaluate {
     publishing {
         publications {
             maven(MavenPublication) {
-                artifactId 'Automattic-Tracks-Android'
-                description 'Android client for Nosara tracks (event tracking and analytics)'
                 from components.release
+
                 groupId 'com.automattic'
+                artifactId 'Automattic-Tracks-Android'
+                artifact tasks.named("androidSourcesJar")
+                // version is set by 'publish-to-s3' plugin
+
+                description 'Android client for Nosara tracks (event tracking and analytics)'
                 pom {
                     licenses {
                         license {
--- a/AutomatticTracks/build.gradle
+++ b/AutomatticTracks/build.gradle
@@ -67,7 +67,6 @@ afterEvaluate {
 
                 groupId 'com.automattic'
                 artifactId 'Automattic-Tracks-Android'
-                artifact tasks.named("androidSourcesJar")
                 // version is set by 'publish-to-s3' plugin
 
                 description 'Android client for Nosara tracks (event tracking and analytics)'
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Automattic-Tracks-Android [![CircleCI](https://circleci.com/gh/Automattic/Automattic-Tracks-Android.svg?style=shield)](https://app.circleci.com/pipelines/github/Automattic/Automattic-Tracks-Android) [![Releases](https://img.shields.io/github/v/release/Automattic/Automattic-Tracks-Android)](https://github.com/Automattic/Automattic-Tracks-Android/releases)
+# Automattic-Tracks-Android
 Client library for tracking user events for later analysis
 
 ## Introduction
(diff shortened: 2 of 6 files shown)


AutomatticTracks/build.gradle
```diff
@@ -63,10 +63,13 @@ afterEvaluate {
...
-                description 'Android client for Nosara tracks (event tracking and analytics)'
                 from components.release
+
                 groupId 'com.automattic'
+                artifactId 'Automattic-Tracks-Android'
+                // version is set by 'publish-to-s3' plugin
+
+                description 'Android client for Nosara tracks (event tracking and analytics)'
                 pom {
                     licenses {
```
oguzkocer, line 74: We don't typically have `description` or `licenses` sections in our POMs. I am not sure what would be the best course of action in this case, so I kept the details of this publication but didn't add the same sections to the `experimentation` module to follow our other projects.

We could consider adding these sections to all projects, but I don't see the need for that yet. I could be convinced otherwise 😅

README.md
```diff
@@ -9,17 +9,35 @@ storing them locally, and on a schedule send them out to the Automattic
 servers. Realistically this library is only useful for Automattic-based
 projects but the idea is to share what we've made.
 
-## Build
-
-* Build:
```
oguzkocer, line 14: I don't think this section has much added value to be part of our `README`, so I wanted to suggest of its removal. This is because `./gradlew assemble` is one of the most basic Gradle tasks and works for all projects.

AutomatticTracks/build.gradle
mokagio: > We could consider adding these sections to all projects, but I don't see the need for that yet. 

+1

This is open source code that anyone can access. Having a description would be helpful and good open source behavior.

At the same time, it's highly unlikely anyone outside Automattic would use our open source tools, because they're tailored to our usage. E.g. To use this library one would need to have their own Tracks instance running and I don't even know if there's a way to configure the base URL to which the library sends event.

Because of that, I'd be leaning towards focusing on more impactful work than adding descriptions to our projects.

settings.gradle
```diff
@@ -1,8 +1,16 @@
 pluginManagement {
     plugins {
         id 'org.jetbrains.kotlin.android' version '1.4.31'
+        id "com.automattic.android.publish-to-s3" version "0.7.0"
```
mokagio: Annoying nitpick. The line above uses single quotes 😅 
```suggestion
        id 'com.automattic.android.publish-to-s3' version '0.7.0'
```

settings.gradle
```diff
@@ -1,8 +1,16 @@
...
     }
     repositories {
+        maven {
+            url 'https://a8c-libs.s3.amazonaws.com/android' 
+            content {
+                includeGroup "com.automattic.android"
+                includeGroup "com.automattic.android.publish-to-s3"
```
mokagio: Another annoying nitpick about quotes usage

```suggestion
                includeGroup 'com.automattic.android'
                includeGroup 'com.automattic.android.publish-to-s3'
```

mokagio (approved):
> As long as Buildkite is green, we should be good for this PR. If there are any issues during usage, which is not very likely, I'll follow up with it as a follow up PR.

Buildkite is green. :shipit:

AutomatticTracks/build.gradle
oguzkocer: Just to add a little more context about this:

I might be mistaken in this, but I assume the description in the POM file is used by maven repositories (such as `mavenCentral`) so it's available for the project page in their site. We don't have such a situation, because we use our own Maven repository and we don't have project pages. So, I don't think it's actually useful to have the description or the license in the POM file.

Having said that, if and when we realize it's useful for something particular, it'll be easier to tailor these details to match that need.

PR #121, opened 2022-03-21
Refactor Experiment so it doesn't depend on ExPlat

### LINKED ISSUE #120: Refactor Experiment so it doesn't depend on ExPlat
We need to refactor the implementation of `Experiment` so it doesn't depend on `ExPlat`. This will help us to avoid having circular dependency problems in the clients (`ExPlat` depends on `Set<Experiment>` and `Experiment` depends on `ExPlat`).

RenanLukas:
Resolves #120 

We need to refactor the implementation of `Experiment` so it doesn't depend on `ExPlat`. This will help us to avoid having circular dependency problems in the clients (`ExPlat` depends on `Set<Experiment>` and `Experiment` depends on `ExPlat`).


experimentation/build.gradle
```diff
@@ -45,6 +46,10 @@ dependencies {
     //WordPress utils
     implementation "org.wordpress:utils:2.3.0"
 
+    //Dagger
+    implementation "com.google.dagger:dagger:2.41"
+    kapt "com.google.dagger:dagger-compiler:2.41"
```
wzieba: 👋 I'm concerned about using `Dagger` in such small library. I understand the purpose of using `Lazy` but I don't think we need to use such powerful tool in this case.

Also it feels dangerous to me: if I'm not mistaken, we loose compile-time validation when injecting `Experiment`s into DI graph. E.g. we can add an `Experiment` in scope of `Fragment`, while `ExPlat` has been created in scope of `Application`. In this case I believe we'll have only `Experiment`s that were created during time of creating `Application` (`Lazy#get` [will refer to the initially injected set](https://dagger.dev/api/2.13/dagger/Lazy.html)).

While it's fixable (inject all experiments in the same scope as `ExPlat`), I think it makes API more difficult and tricky to work with.

---

I don't believe that `Experiment` needs `ExPlat` dependency. WDYT about:

```kotlin
// If I understand correctly, this class is meant to be inherited from
open class Experiment(
    val name: String,
)
```

and then in `ExPlat.kt`

```kotlin 
fun getVariation(experiment: Experiment, shouldRefreshIfStale: Boolean): Variation
```

WDYT?

renanferrari: FWIW I think these are valid concerns. The original implementation using `Lazy` wasn't ideal – even considering it wasn't being used in a library.

If I'm not mistaken, the idea of having `ExPlat` as a dependency of `Experiment` was to make it clearer to understand where each experiment was being used, as a class would explicitly declare its dependency on an individual experiment instead of `ExPlat`. We can probably achieve a similar result by refactoring `ExPlat` with a better separation of responsibilities, but the way I see it, this is more of a nice-to-have than a must-have, so I agree with the proposal of simply removing the `ExPlat` dependency from `Experiment`.

One thing I want to add is that if we decide to go down this route, I would also consider making `Experiment` an interface instead of an open class.

RenanLukas:
Hey @wzieba and @renanferrari , thanks for taking the time to review my PR 👍 

I've updated it following your suggestions.

experimentation/src/main/java/com/automattic/android/experimentation/Experiment.kt
```diff
@@ -1,17 +1,5 @@
...
-    @Suppress("unused")
-    @JvmOverloads
-    fun getVariation(shouldRefreshIfStale: Boolean = false): Variation {
-        return exPlat.getVariation(
-            experimentName = name,
-            shouldRefreshIfStale = shouldRefreshIfStale
-        )
-    }
+interface Experiment {
+    fun name(): String
```
renanferrari: We should probably keep this a `val`, as the experiment name shouldn't change.

renanferrari (commented):
Good job on this! I left a minor suggestion below, but other than that this looks pretty good to me 👍

Also, since we're not using `Lazy` anymore, you should probably update the PR name to reflect the actual changes.

experimentation/build.gradle
```diff
@@ -1,13 +1,14 @@
 plugins {
-    id 'com.android.library'
-    id 'org.jetbrains.kotlin.android'
+    id "com.android.library"
+    id "org.jetbrains.kotlin.android"
+    id "org.jetbrains.kotlin.kapt"
```
wzieba: If I'm not mistaken, as we don't use Dagger here anymore, we can remove `kapt` plugin.

```suggestion
```

experimentation/src/main/java/com/automattic/android/experimentation/Experiment.kt
RenanLukas: Good point 👍 
Done

experimentation/build.gradle
RenanLukas: Ops! Forgot about this one 😅 
Removed, thanks!

wzieba (approved):
Looks good to me, thanks for improvements!

renanferrari (approved):
Thank you, the changes are looking good! I'm approving this PR, but I also wanted to ask something for you and @wzieba: how do you plan to provide the experiments to `ExPlat`?

On the WordPress side, I think things could be simplified a lot by using an `enum` that extends from `Experiment` like the following:

```kotlin
enum class WordPressExperiment(override val name: String) : Experiment {
    EXAMPLE_TEST_AB("example_experiment"),
    LAND_ON_THE_EDITOR_AB("wpandroid_land_in_the_editor_phase1_v3");
}
```

Then using Dagger to create an `ExPlat` instance would look something like this:

```kotlin
@Provides
@Singleton
fun provideExPlat(
    experimentStore: ExperimentStore,
    appLogWrapper: AppLogWrapper,
    @Named(APPLICATION_SCOPE) coroutineScope: CoroutineScope
) = ExPlat(
        Platform.WORDPRESS_ANDROID,
        WordPressExperiment.values().toSet(),
        experimentStore,
        appLogWrapper,
        coroutineScope,
        BuildConfig.DEBUG
)
```

And usage would be as simple as: 

```kotlin
class SomeClass @Inject constructor(val exPlat: ExPlat) {
    fun doSomething() {
        when (exPlat.getVariation(WordPressExperiment.SOME_EXPERIMENT)) {
            is Control -> // Do something
            else -> // Do something else
        }
    }
}
```

However, I just noticed that this is not possible because the `Experiment.name` property clashes with `enum.name`. This can be solved by renaming the property to something else, but before I suggested that, I just wanted to check if you folks had another approach in mind. What do you say?

RenanLukas:
@renanferrari that's a very good point! I agree with renaming the field to something else (I'm just not sure what should we call it 😂 ). A few suggestions, please feel free to suggest other names:
`identifier`
`constant`
`value`

What do you both think?

Also, if you have a different suggestion (other than renaming), please let me know.

RenanLukas:
I've renamed it to `identifier` for now 👍

Issue #102, opened 2021-07-19
Extract `crashloggin` package to a separate module


wzieba:
See internal discussion: p1626707401400500/1626676695.387300-slack-CC7L49W13

PR #135, opened 2022-07-20
Update Gradle and AGP

wzieba:
### Description

As part of maintenance efforts and to allow efficient work, I propose a set of updates for the project.

This PR updates:
- Gradle Wrapper
- Android Gradle Plugin
- Target and Compile SDKs


AutomatticTracks/build.gradle
```diff
@@ -20,23 +20,22 @@ dependencies {
...
+    compileSdkVersion 33
 
     defaultConfig {
-        versionName "2.1.0"
         minSdkVersion 21
-        //noinspection OldTargetApi
-        targetSdkVersion 30
+        targetSdkVersion 33
         buildConfigField "String", "SENTRY_TEST_PROJECT_DSN", project.properties.get("sentryTestProjectDSN")
+        buildConfigField "String", "VERSION_NAME", "\"2.1.0\""
```
wzieba, line 38: It was a constant before, I'm not changing anything here. But it's invalid. More details internally: pdnsEh-IK-p2

AutomatticTracks/build.gradle
```diff
@@ -20,23 +20,22 @@ dependencies {
 
     testImplementation 'junit:junit:4.13.2'
     testImplementation 'org.assertj:assertj-core:3.19.0'
-    testImplementation 'org.mockito.kotlin:mockito-kotlin:2.2.11'
-    testImplementation 'org.mockito:mockito-inline:3.8.0'
+    testImplementation 'org.mockito.kotlin:mockito-kotlin:4.0.0'
+    testImplementation 'org.mockito:mockito-inline:4.6.1'
```
ParaskP7: Question (❓): Why did you decide to update to this version of `mockito-inline` and not its latest version? FYI: I am seeing that the latest version of this library is the [4.8.1](https://mvnrepository.com/artifact/org.mockito/mockito-inline).

PS: The same question (❓) applies to both, the `junit` and `runner` updates below, which are related to [androidx.test](https://developer.android.com/jetpack/androidx/releases/test), with their latest versions being the [1.1.4](https://developer.android.com/jetpack/androidx/releases/test#ext.junit-1.1.4) and [1.5.0](https://developer.android.com/jetpack/androidx/releases/test#runner-1.5.0) respectively.

AutomatticTracks/build.gradle
ParaskP7: Praise (❤️): Yeap, yeap, thank you for this change and for providing more details on that. Library modules no longer have the `versionName/Cope` configuration, only application modules have that with the latest AGP.

AutomatticTracks/build.gradle
```diff
@@ -20,23 +20,22 @@ dependencies {
...
 android {
-    compileSdkVersion 30
+    compileSdkVersion 33
 
     defaultConfig {
-        versionName "2.1.0"
         minSdkVersion 21
-        //noinspection OldTargetApi
-        targetSdkVersion 30
+        targetSdkVersion 33
```
ParaskP7, line 36: 1. Suggestion (💡): Would you consider taking this opportunity and extracting these `minSdkVersion`, `compileSdkVersion` and `targetSdkVersion` into `settings.gradle` so that to make sure that when we next upgrade those we do it once and for all, just like it is already done for WP/WCAndroid?

`settings.gradle`
```
gradle.ext {
    minSdkVersion = 21
    compileSdkVersion = 33
    targetSdkVersion = 33
}
```

`build.gradle`
```
    compileSdkVersion gradle.ext.compileSdkVersion
    defaultConfig {
        minSdkVersion gradle.ext.minSdkVersion
        targetSdkVersion gradle.ext.targetSdkVersion
    }
```

2. Suggestion (💡): On the same spirit, could you please consider doing the same for at least the `test` dependencies that you updated, which was also done multiple times, making it more error prone that it could have been if that was done once and for all. PS: Feel free to do it for `all` the dependencies as well, if that is not to much of a hustle for you.

experimentation/build.gradle
```diff
@@ -32,7 +32,12 @@ android {
     }
     lintOptions {
         warningsAsErrors true
-        warning 'GradleDependency'
```
ParaskP7, line 35: Minor (🔍): I think one of those was missed from being deleted as well. See the `lintOptions { ... }` configuration on the `sampletracksapp` app module.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.1.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.5.1-all.zip
```
ParaskP7, line 4: Suggestion (💡): How about taking this opportunity and introduce `distributionSha256Sum` here as well (just like it was done recently for the [pocket-casts-android](https://github.com/Automattic/pocket-casts-android) repo (see code [here](https://github.com/Automattic/pocket-casts-android/blob/main/gradle/wrapper/gradle-wrapper.properties#L7) and PR [here](https://github.com/Automattic/pocket-casts-android/pull/419)).

Thus, what you could do is:
1. Find the correct [complete (-all) zip checksum](https://gradle.org/release-checksums/) for this `7.5.1` version of Gradle.
2. Run the below command after having copy-pasted the correct checksum into the `gradle-distribution-sha256-sum` option:
 
`./gradlew wrapper --gradle-version=7.5.1 --distribution-type=all --gradle-distribution-sha256-sum=db9c8211ed63f61f60292c69e80d89196f9eb36665e369e7f00ac4cc841c2219`

settings.gradle
```diff
@@ -1,5 +1,6 @@
 pluginManagement {
     plugins {
+        id 'com.android.library' version '7.3.0'
```
ParaskP7: Suggestion (💡): Consider updating to the latest [7.3.1](https://maven.google.com/web/index.html?q=gradle#com.android.tools.build:gradle:7.3.1) version of `com.android.library`, that is of course, unless you found an issue and couldn't do that in the first place.

PS: You might want to update `lint-baseline.xml` again after applying this change (see `by="lint 7.3.0"`).

ParaskP7 (commented):
👋 @wzieba !

I have reviewed this PR, everything LGTM, thank you so much for taking such good care of this repo! 🌟 ❤️ 🙇 

ℹ️ I've also tested this PR using the `SampleTracksApp` and sending events to my personal sample test Sentry project. However, I only did verify that by manually testing it. FYI: For some unknown reason, I couldn't trigger the `SendEventsToSentry` automated tests, but I also didn't want to spend more time on troubleshooting that as you might be aware why this wasn't working. PS: Does it work for you after this upgrade? 🤔

-----

I have left few questions (❓), a couple of suggestions (💡) and some minor (🔍) comments for you to consider. I am **NOT** going to approve this PR for now, but nothing is really blocking. I just want to give you some time to apply any of my suggestions before merging this to `trunk`.

AutomatticTracks/build.gradle
wzieba: >  Why did you decide to update to this version of mockito-inline and not its latest version?

I created this PR some time ago (20th Jul) and then it was probably the newest version 😅 . Thanks for checking this!

AutomatticTracks/build.gradle
wzieba: Thanks for leaving those suggestions! I'd opt not to centralize the SDK and/or dependencies versioning. Each module in this project is independent - `AutomatticTracks` and `experimentation` can be integrated into projects without each other, and they don't influence each other. They leave in the same repository just because of pragmatism.

`sampleproject` is just a test app, I don't think we should enforce strictly the same version of SDK or dependencies between them.

It's contrary to WP/WCAndroid apps where modules depend on each other and syncing versions is indeed beneficial. But for Automattic-Tracks-Android, I think that fewer dependencies is a better approach.

wzieba:
Thanks @ParaskP7 for insightful, like always, review! I've implemented majority of your suggestions.

> However, I only did verify that by manually testing it. FYI: For some unknown reason, I couldn't trigger the SendEventsToSentry automated tests

Sorry for not mentioning it - I'll be removing `SendEventsToSentry` test. I discussed it internally here with many details: pdnsEh-vG-p2#comment-1067 . So I didn't even check if this test works 😅 - we can safely ignore it.

AutomatticTracks/build.gradle
ParaskP7: Thanks for your comment @wzieba ! 🙇 

> I'd opt not to centralize the SDK and/or dependencies versioning. Each module in this project is independent - AutomatticTracks and experimentation can be integrated into projects without each other, and they don't influence each other.

Yes, this totally makes sense, thank you for thinking about that and for deciding not to go with that centralization approach for the reason you wrote. 👍 

However, me looking at this 125f7b704ddea040e3d588ec05f6aa8351ba667c of yours and the fact that all those SDK versions are exactly the same, this begs the question, why not keeping it as centralized, at least for now, and as much as possible, especially if it is already like that. Then, consequent upgrades can continue happening in a centralized manner and only de-centralize in specific and unique situations, when needed, and in isolation? 🤔

> They leave in the same repository just because of pragmatism.

Because I am not fully aware of what the relation is between `AutomatticTracks`, `experimentation` and `sampletracksapp`, and since it seems that you are, then maybe this is a good opportunity to make it clearer for all developers, that is, what is what. So, to my understanding:

`sampletracksapp` is using `AutomatticTracks` and as such those could be related and their SDK centralized.
`experimentation` seems to be an isolated library and not being used in either `AutomatticTracks` or via `sampletracksapp`, thus this can be de-centralized like that. Doing so and making this obvious will help others read this repo a lot faster. 🤔

So, all-in-all I am seeing the issue being `experimentation` and not `sampletracksapp` + `AutomatticTracks` becoming centralized. Correct me if I am wrong. 🙏 

> sampleproject is just a test app, I don't think we should enforce strictly the same version of SDK or dependencies between them.
> 
> It's contrary to WP/WCAndroid apps where modules depend on each other and syncing versions is indeed beneficial. But for Automattic-Tracks-Android, I think that fewer dependencies is a better approach.

Yes, 100%, but then again, look at what happened to `FluxC`, the example app was on `targetSdkVersion = 22` for so long and it now becomes a blocker to even update its `minSdkVersion` (see [here](https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2563)).

I actually consider the sample apps to be on the same boat with WP/WCAndroid and would vote in favor of treating those as such, just so that we are aware the problems with doing such SDK updates on the library level first, via their example apps and then move forward with the SDK update on the clients side. I really find lots of benefits to following such a "centralized" approach.

Obviously, as I tried to explain above, I am also in favor of "de-centralizing" but only when it is really needed (ie `experimentation`. Else, I would recommend to put that bit of an extra effort to upgrade even the example apps, then test those, and only leave them be if they are blocking the upgrade of the individual libraries by a significant amount.

No matter, leaving any library or client or example app behind adds to tech debt that is going to be pained anyway overtime, thus by paying just a little bit of tech debt progressively feels a lot more maintainable to me.

-----

PS: The above is just food for thought, please don't feel the need to action on any of what I am saying to you, apply an changes, yara yara yara... This is just us developers discussing! 😄

ParaskP7:
> Thanks @ParaskP7 for insightful, like always, review! I've implemented majority of your suggestions.

Thank YOU, for applying any of my comments and for the thoughtful discussions, as always, a pleasure doing business with you! ❤️ 

> Sorry for not mentioning it - I'll be removing SendEventsToSentry test. I discussed it internally here with many details: pdnsEh-vG-p2#comment-1067 . So I didn't even check if this test works 😅 - we can safely ignore it.

Ah, thank you, I am ignoring it then! 💯

ParaskP7 (approved):
👋 @wzieba !

I reviewed and tested the PR update, everything LGTM, thank you so much for working on this, you rock, feel free to merge this at your convenience! 🚀

AutomatticTracks/build.gradle
```diff
@@ -11,32 +11,31 @@ repositories {
 
 dependencies {
     implementation platform('io.sentry:sentry-bom:6.3.0')
-    implementation "io.sentry:sentry-android"
-    implementation "io.sentry:sentry-android-okhttp"
-    implementation "io.sentry:sentry-android-fragment"
+    implementation 'io.sentry:sentry-android'
+    implementation 'io.sentry:sentry-android-okhttp'
+    implementation 'io.sentry:sentry-android-fragment'
```
ParaskP7: Suggestion (💡: Btw, a quick FYI for your @wzieba . I am always in favor of having double quotes (`"`) instead of single quotes (`'`) like that.

This is twofold and because:
1. You get a warning/error (a 🔴 underline at least) like that if you try to use a syntax like `'androidx.annotation:annotation:${version}'`, not the `${version}`, see here: `It looks like you are trying to substitute a version variable, but using single quotes ('). For Groovy string interpolation you must use double quotes (").` 
2. If we were to ever migration to Kotlin Gradle DSL, this would become a prerequisite, to have double quotes (`"`) everywhere.

wzieba: Thanks, moved all of them to `"`!

AutomatticTracks/build.gradle
wzieba: > this begs the question, why not keeping it as centralized, at least for now, and as much as possible, especially if it is already like that

That's a good point, maybe we can centralize and decentralize when (or _if_) it'll be indeed?

> Because I am not fully aware of what the relation is between AutomatticTracks, experimentation and sampletracksapp, and since it seems that you are, then maybe this is a good opportunity to make it clearer for all developers (...)

That's a good point, I think we should clarify it. The knowledge I have is from internal P2s. I see an opportunity for clarifying relations in the scope of #102

wzieba:
Thanks @ParaskP7 🙏 !

AutomatticTracks/build.gradle
ParaskP7: > That's a good point, maybe we can centralize and decentralize when (or if) it'll be indeed?

Yeap, maybe! 😅 

> That's a good point, I think we should clarify it. The knowledge I have is from internal P2s. I see an opportunity for clarifying relations in the scope of https://github.com/Automattic/Automattic-Tracks-Android/issues/102

Cool, right, this #102 seems a good opportunity to do so indeed! 💯

PR #136, opened 2022-07-22
Sentry Performance configuration

wzieba:
## Description

This PR adds support for the Sentry Performance Monitoring feature for clients with an anti-bikeshading approach. We expose only limited portion of easy to understand API.

As our needs will grow in the future, we'll expose more needed API. But for now it's completely sufficient. 

### How to test
1. Have a test Sentry project with Performance Monitoring enabled.
2. Launch the test app
3. Click on `Measure performance of sample HTTP fetch`
4. Make sure that after a while you can see a report similar to:

image


AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/RequestFormatter.kt
```diff
@@ -0,0 +1,14 @@
...
+import okhttp3.Request
+
+/**
+ * Formats OkHttp requests before using them as a breadcrumb or performance span
+ *
+ * @return formatted url, sent to Sentry
+ */
+interface RequestFormatter {
+    fun formatRequestUrl(request: Request): FormattedUrl
+}
```
wzieba, line 12: We want to format the URL to:
- obfuscate it from private data
- make merging the same endpoints easier in Sentry

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryCrashLogging.kt
```diff
@@ -1,37 +1,56 @@
...
+                    when (val perfConfig = dataProvider.performanceMonitoringConfig) {
+                        Disabled -> null
+                        is Enabled -> perfConfig.sampleRate
+                    }
                 isDebug = dataProvider.enableCrashLoggingLogs
                 setTag("locale", dataProvider.locale?.language ?: "unknown")
-                beforeSend = SentryOptions.BeforeSendCallback { event, _ ->
+                setBeforeBreadcrumb { breadcrumb, _ ->
+                    if (breadcrumb.type == "http") null else breadcrumb
+                }
```
wzieba, line 43: We don't want to add HTTP breadcrumbs for privacy reasons.

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/performance/PerformanceMonitoringRepositoryProvider.kt
```diff
@@ -0,0 +1,10 @@
+package com.automattic.android.tracks.crashlogging.performance
+
+import com.automattic.android.tracks.crashlogging.performance.internal.SentryPerformanceMonitoringWrapper
+
+object PerformanceMonitoringRepositoryProvider {
+
+    private val instance = PerformanceTransactionRepository(SentryPerformanceMonitoringWrapper())
+
+    fun createInstance(): PerformanceTransactionRepository = instance
+}
```
wzieba, line 10: This makes `PerformanceTransactionRepository` singleton.

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/performance/internal/SentryPerformanceMonitoringWrapper.kt
```diff
@@ -0,0 +1,11 @@
+package com.automattic.android.tracks.crashlogging.performance.internal
+
+import io.sentry.ITransaction
+import io.sentry.Sentry
+
+class SentryPerformanceMonitoringWrapper {
```
wzieba, line 6: This class makes `PerformanceTransactionRepository` testable. We couldn't mock direct `Sentry.startTransaction` otherwise.

AutomatticTracks/src/main/AndroidManifest.xml
```diff
@@ -9,5 +9,7 @@
         <meta-data
             android:name="io.sentry.auto-init"
             android:value="false" />
+
+        <meta-data android:name="io.sentry.breadcrumbs.user-interaction" android:value="false" />
```
wzieba, line 13: Disables a new (enabled by default) feature of Sentry to log every UI interaction like this:
Screenshot 2022-08-02 at 11 18 46

ThomazFB (approved):
Such a fine work here, @wzieba 💯, nice implementation strategy and great awareness of our users privacy. Thank you for this careful implementation! Look good to me, but I'll leave it just approved in case you want a second pair of eyes here.

AutomatticTracks/src/main/AndroidManifest.xml
ThomazFB: I wonder if it could be useful to use this UI interaction breadcrumbs, but I imagine it can be too overwhelming depending on the scenario. Was that the reason to disable it?

wzieba: I decided to disable it (as almost any other breadcrumb: HTTP, fragment lifecycle) because of privacy concerns (internal link: p91TBi-5Bu-p2)

wzieba:
Thanks @ThomazFB for the review, I think we're good to go with merging it!

PR #137, opened 2022-07-28
Observe user and app context changes

wzieba:
## Description

This PR changes an approach of assigning user and application context (tags) to Sentry events.

### Why?

The currently used (a little hacky) approach does not work when uploading Sentry Performance traces. The traces are uploaded without information about users and site (app context).

#### Current approach

image

Before sending every event, the `BeforeSendCallback` is executed. In this callback, we request user information from the app via `CrashLoggingDataProvider`, which takes the user from e.g. an `AccountStore`.

This approach does not work when uploading Sentry Performance Monitoring Traces. I don't know the exact reason but, as it's against the framework, I've decided to change our approach. [Here are docs about setting the user.](https://docs.sentry.io/platforms/android/enriching-events/identify-user/)

#### The new approach

image

Instead of sequentially requesting user and app context before sending every event, we [observe](https://github.com/Automattic/Automattic-Tracks-Android/pull/137/files#diff-9b71c469119566ba0b57a20395609bef014d9604b43fb253e9590cfb6c86559bR42-R51) new data as it comes.

### How to test

1. Please open `SendEventsToSentry.kt` test. Don't worry about IDE underlining code with red. It's known, "won't fix" issue of IDE.
2. Read the instruction at the top.
3. Every test is self-described, so you should be good to go. Tip: I suggest removing all events in your test project beforehand for easier navigation
4. In case of trouble feel free to ping me.


AutomatticTracks/build.gradle
```diff
@@ -15,11 +15,13 @@ dependencies {
     implementation "io.sentry:sentry-android-okhttp"
     implementation 'androidx.annotation:annotation:1.1.0'
     implementation 'com.squareup.okhttp3:okhttp:4.9.0'
+    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.6.4'
```
wzieba, line 18: Both cliensts user coroutines. I found this approach more effective and readable than custom listeners/emmiters.

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryCrashLogging.kt
```diff
@@ -23,28 +26,29 @@ internal class SentryCrashLogging constructor(
...
+        applicationScope.launch {
+            dataProvider.user.collect {
+                sentryWrapper.setUser(it?.toSentryUser())
+            }
+        }
+        applicationScope.launch {
+            dataProvider.applicationContextProvider.collect {
+                sentryWrapper.setTags(it)
+            }
+        }
```
wzieba, line 53: Hotspot of the PR: here, we observe incoming new data and update Sentry, independently from `SentryOptions.BeforeSendCallback` callback.

AutomatticTracks/src/sharedTest/kotlin/com/automattic/android/tracks/fakes/FakeDataProvider.kt
```diff
@@ -13,17 +15,21 @@ class FakeDataProvider(
...
-    override fun userProvider(): CrashLoggingUser? {
-        return user
-    }
+    val fakeUserEmitter = MutableStateFlow(initialUser)
+
+    val fakeApplicationContextEmitter = MutableStateFlow(initialApplicationContext)
+
+    override val user: Flow<CrashLoggingUser?> = fakeUserEmitter
+
+    override val applicationContextProvider: Flow<Map<String, String>> = fakeApplicationContextEmitter
```
wzieba, line 32: You can see a similar approach in `FakeDataSource` in [Android Testing docs](https://developer.android.com/kotlin/flow/test#continuous-collection)

AutomatticTracks/src/test/java/com/automattic/android/tracks/crashlogging/SentryCrashLoggingTest.kt
```diff
@@ -51,13 +61,14 @@ class SentryCrashLoggingTest {
...
             provideExtrasForEvent = provideExtrasForEvent,
-            applicationContext = applicationContext,
+            initialApplicationContext = applicationContext,
         )
 
         crashLogging = SentryCrashLogging(
             context = mockedContext,
             dataProvider = dataProvider,
             sentryWrapper = mockedWrapper,
+            applicationScope = testScope.backgroundScope
```
wzieba, line 71: From the docs: 
image

It seems like a perfect use case for us as `SentryCrashLogging` starts 2 coroutines in its constructor (which don't stop).

ThomazFB (approved):
Nice improvement, @wzieba, and sorry for the long time to review this PR! It was my first one in this repo 😅  I like the solution of using a reactive approach of the User data to configure Sentry's current user. LGTM!

PR #138, opened 2022-07-29
Update `ktlint` to 0.46.1. Fix newly reported issues

wzieba:
Ktlint in version `0.40.0` is a blocker for a few Kotlin features, e.g. it doesn't allow to pass `value class`, reporting an error. This is why I want to update this tool.

### Trailing comas
Ktlin docs: https://pinterest.github.io/ktlint/rules/configuration/#trailing-comma

Trailing commas in Ktlint are disabled by default. If we enable them, the trailing commas start to be forced to every possible place.

I find it kinda odd, but as Ktlint is designed to be anti-bikeshedding, I'd like to align with default rules. I don't think trailing commas are something we should care about in this small library.


wzieba:
hi @ParaskP7 👋 

Are you okay with reviewing this PR? It's a pure configuration one. I only updated ktlint, moved 2 methods and run `./gradlew ktlintFormat`.

ParaskP7:
👋 @wzieba !

> Are you okay with reviewing this PR? It's a pure configuration one. I only updated ktlint, moved 2 methods and run ./gradlew ktlintFormat.

100%, I'll be reviewing this PR today, thanks for the ping! ❤️

ParaskP7 (approved):
👋 @wzieba !

I reviewed and testing this PR, everything LGTM! 🥇 

PS: The only minor thing I would recommend for next time is to use a separate commit for when it comes to refactoring code (like moving the `appendTags(...)` and `toSentryUser(...)` functions), or in general, when applying an unrelated code change to a specific solution related commit.

wzieba:
Thanks @ParaskP7 ! I'll keep that in mind 👍

PR #40, opened 2019-01-18
Log device orientation + bump version to 1.1.5

malinajirka:
Fixes #39

- adds `google()` repository to build.gradle
- adds device orientation property to immutableProperties
- bumps version to 1.1.5

`getConfiguration().orientation` may get out of sync. Therefore we use it only when the app starts to find out whether the natural (default) device orientation is portrait/landscape. The current device orientation is derived from `rotation`.

To test:
I'm not sure whether there is an easy way how to test this. I've added a log to DeviceInformation and published an artifact to local maven repository.

Note: We need to publish the new artifact to bintray.

PR #141, opened 2022-08-29
Add tracking for font-sizing related info.

hafizrahman:
To solve https://github.com/woocommerce/woocommerce-android/issues/7282

This adds the following data:

- fontScale
- scaledDensity

The goal is so we know how many users set the font super high and be affected by display oddities with larger fonts.

I'm not yet sure how different `fontScale` and `scaledDensity` values will be on different type of screen out there, and whether one or the other is enough to determine how big the font is set by the user. So I thought it'd be safer to get both.

I am not sure how to test this but previous PRs like https://github.com/Automattic/Automattic-Tracks-Android/pull/40 that does similar thing does not seem to include any.

@wzieba if I understand correctly from the `build.gradle` file in WCAndroid, we just use `com.automattic:Automattic-Tracks-Android:trunk-<commit full SHA1>`. So there's nothing specific here to do after this is PR is merged, just add the change in WCAndroid?

### DIFF
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/DeviceInformation.java
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/DeviceInformation.java
@@ -185,6 +185,18 @@ import java.util.Locale;
             Log.e(LOGTAG, "Exception writing display_density_dpi value in JSON object", e);
         }
         try {
+            // Track font-sizing related values.
+            // - fontScale: user preference for scaling factor. See: https://developer.android.com/reference/android/content/res/Configuration.html#fontScale
+            // - scaledDensity: density * fontScale. See: http://androidxref.com/4.2_r1/xref/packages/apps/Settings/src/com/android/settings/Display.java#99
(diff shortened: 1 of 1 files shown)


wzieba:
> @wzieba if I understand correctly from the build.gradle file in WCAndroid, we just use com.automattic:Automattic-Tracks-Android:trunk-<commit full SHA1>. So there's nothing specific here to do after this is PR is merged, just add the change in WCAndroid?

Yes, exactly.

_More context_
But I think we should eventually change this some day and use semantic versioning. Tracks library is not updated that often to justify FluxC-like hash versioning. It might be helpful to get a context of changes quickly. We also didn't change versioning because we wanted to - it was just a kind of a side effect of moving to our internal repository.

JorgeMucientes:
Looks good @hafizrahman thanks for dealing with it. Curious about the results coming out of this. 
🚢

JorgeMucientes:
Looks good @hafizrahman thanks for dealing with it. Curious about the results coming out of this. 
🚢

JorgeMucientes (approved):

PR #143, opened 2022-09-15
Provide transaction sampler

wzieba:
### What?

Introduce performance transaction sampler that allows to provide different sampling values depending on the transaction type.

### Why?

E.g. we want to measure every 100th event of opening Order Details (because it often happens) but we would like to measure every single site creation process (because it happens rarely)


ParaskP7 (approved):
👋 @wzieba !

I have reviewed the changes included in this PR, but not tested it as I am not sure how to test this other by running the `SentryCrashLoggingTest` unit test. 🤷 

All the changes LGTM, thus, I am going to approve this PR and not block you, especially since this is just a child PR of this other [parent PR](https://github.com/Automattic/Automattic-Tracks-Android/pull/172). However, do let me know if I can somehow test it myself too, and maybe what's the best/quickest way (if any). Also, maybe you also can reference how the `CrashLoggingDataProvider` + `PerformanceSampler` is to be used on WCAndroid, maybe that will give me a better idea. 🤔

wzieba:
hi @ParaskP7, thank you for the review!

> Also, maybe you also can reference how the CrashLoggingDataProvider + PerformanceSampler is to be used on WCAndroid, maybe that will give me a better idea. 🤔

Good idea, I'll prepare PR in WCAndroid to showcase the usage. I'm not merging this PR for now then.

ParaskP7:
👋 @wzieba !

> Good idea, I'll prepare PR in WCAndroid to showcase the usage.

Awesome! 🥇 

> I'm not merging this PR for now then.

I am actually comfortable for you to merge this PR, even now. Then, we can make any refinement on the other [parent PR](https://github.com/Automattic/Automattic-Tracks-Android/pull/172). It is totally up to you how to want to move forward. 💯

wzieba:
It's not that needed anymore. I'll close this PR for now, we can always return to it.

PR #145, opened 2022-09-28
Add user type for Pocket Casts

ashiagr:
This adds user type for Pocket Casts app.

Uses whitelisted user type added here:
https://github.com/Automattic/nosara/commit/01bccdfbfcb650f3577fa4ba95093dfc4c23ea20


mchowning (approved):
Tested via the related Pocket Casts PR and this is working well. Nice job!

PR #167, opened 2023-03-15
Add user type for Day One

khaykov:
We are adding Tracks to Day One Android application. This PR adds a corresponding user type.

`dayone:user_id` is already whitelisted  [here](https://github.com/Automattic/nosara/blob/master/wyeast_streaming/src/main/resources/tracks/whitelists/tracks_useridtype_whitelist.config).

There is nothing to test.


khaykov:
On advice from Danilo, I'll add a client side PR as well, so we can make sure everything is actually working.

danilo04 (approved):
This is working as expected @khaykov. Great job!

PR #188, opened 2023-08-03
Update dependabot.yml to check in main directory

wzieba:
After centralizing version numbers, the version is no longer available directly in AutomatticTracks/build.gradle file, but rather in /build.gradle.

See https://github.com/Automattic/Automattic-Tracks-Android/pull/187#issuecomment-1662492428


ParaskP7 (approved):
LGTM, thanks for making this change @wzieba ! 💯

PR #196, opened 2023-10-24
[Nullability Annotations to Java Classes] Add `wordpress.lint` Dependency

ParaskP7:
This PR adds `wordPress-lint` checks to the project ([2.0.0](https://github.com/wordpress-mobile/WordPress-Lint-Android/releases/tag/2.0.0)).

-----

## To test:

1. Verifying that all the CI checks are successful (focus on the Lint related [check](https://buildkite.com/automattic/automattic-tracks-android/builds/191#018b61d2-63fd-4d59-aa7f-1e00f5bd7235) and its [report](N/A)).
2. Verify that the new `MissingNullAnnotationOn*` correctness related rules are reporting as expected. For a reference, see screenshot below:

image

### DIFF
--- a/AutomatticTracks/build.gradle
+++ b/AutomatticTracks/build.gradle
@@ -11,7 +11,6 @@ repositories {
         url "https://a8c-libs.s3.amazonaws.com/android"
         content {
             includeGroup "org.wordpress"
-            includeGroup "org.wordpress.lint"
         }
     }
 }
--- a/AutomatticTracks/build.gradle
+++ b/AutomatticTracks/build.gradle
@@ -7,6 +7,13 @@ plugins {
 repositories {
     google()
     mavenCentral()
+    maven {
+        url "https://a8c-libs.s3.amazonaws.com/android"
+        content {
+            includeGroup "org.wordpress"
+            includeGroup "org.wordpress.lint"
+        }
+    }
 }
 
 dependencies {
@@ -23,6 +30,8 @@ dependencies {
     testImplementation "org.mockito.kotlin:mockito-kotlin:$mockitoKotlinVersion"
     testImplementation "org.mockito:mockito-inline:$mockitoInlineVersion"
     testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:$kotlinxCoroutinesVersion"
+
+    lintChecks "org.wordpress:lint:$wordpressLintVersion"
 }
 
 android {
--- a/build.gradle
+++ b/build.gradle
@@ -57,4 +57,7 @@ ext {
     junitVersion = '4.13.2'
     mockitoInlineVersion = '4.8.1'
     mockitoKotlinVersion = '4.0.0'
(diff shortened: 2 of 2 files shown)


AutomatticTracks/build.gradle
```diff
@@ -7,6 +7,13 @@ plugins {
 repositories {
     google()
     mavenCentral()
+    maven {
+        url "https://a8c-libs.s3.amazonaws.com/android"
+        content {
+            includeGroup "org.wordpress"
+            includeGroup "org.wordpress.lint"
```
wzieba:
```suggestion
```

The `group` for lint library is `org.wordpress` - there's no need to add the other group (which is not valid AFAIK, `lint` is the name of a module)

wzieba (approved):

AutomatticTracks/build.gradle
ParaskP7: Yeap, I wasn't 100% sure about adding `org.wordpress.lint` too, I just followed the same pattern across all of our other repos, see this [Login](https://github.com/rohankumardubey/WordPress-Login-Flow-Android/blob/3ab114c06b2311e4f25d28adced9f70c34bfbbac/WordPressLoginFlow/build.gradle#L27-L33) repo as an example... 🤷 

Wdyt @wzieba ? 🤔

ParaskP7: Also to make the `example` module work on the `Catalog` repo, I added [this](https://github.com/Automattic/android-dependency-catalog/blob/trunk/example/settings.gradle.kts#L36) line there too, following the [Adding a new dependency](https://github.com/Automattic/android-dependency-catalog/blob/trunk/example/settings.gradle.kts#L36) documentation, as I was indeed hitting a problem on it... 🤷 

> [Build the catalog & example projects](https://github.com/Automattic/android-dependency-catalog#building-locally) - if you get Could not find {dependency} error, make sure the repository for the new dependency you've added is included in [example/settings.gradle.kts](https://github.com/Automattic/android-dependency-catalog/blob/trunk/example/settings.gradle.kts)

wzieba: > Yeap, I wasn't 100% sure about adding org.wordpress.lint too, I just followed the same pattern across all of our other repos, see this Login repo as an example... 🤷

It seems that this part in `Login` is also incorrect - it adds a valid group `org.wordpress` but also invalid and unnecessary one: `org.wordpress.fluxc`.

> Also to make the example module work on the Catalog repo, I added this line there too, following the Adding a new dependency documentation, as I was indeed hitting a problem on it... 🤷

Yes, this line is correct - it adds a new `module` (not `group`) so it specifies `lint` module with it's group: `org.wordpress`.

ParaskP7: Got it, thanks for the explanation @wzieba ! 🥇 

Done: 1e055a63d5253553bd160f3998736df642640512

PR #202, opened 2024-03-06
Add support to send JS exceptions to Sentry

jhnstn:
**Related Changes:** 
- https://github.com/wordpress-mobile/WordPress-Android/pull/20359

This PR expands the Crash logging service to support logging JavaScript exceptions to Sentry. This will allow hybrid sources like React Native (used in Gutenberg Mobile) to log exceptions with detailed information and symbolicated stack traces, which will greatly simplify the crash debugging process.

### To test

Since it involves testing exceptions, we need to modify the code to force them and generate an installable build. A test PR has been created for this purpose, follow the testing instructions from https://github.com/wordpress-mobile/gutenberg-mobile/pull/6654.

### DIFF
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryCrashLogging.kt
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryCrashLogging.kt
@@ -5,6 +5,8 @@ import com.automattic.android.tracks.crashlogging.CrashLogging
 import com.automattic.android.tracks.crashlogging.CrashLoggingDataProvider
 import com.automattic.android.tracks.crashlogging.CrashLoggingUser
 import com.automattic.android.tracks.crashlogging.ExtraKnownKey
+import com.automattic.android.tracks.crashlogging.JsException
+import com.automattic.android.tracks.crashlogging.JsExceptionCallback
 import com.automattic.android.tracks.crashlogging.PerformanceMonitoringConfig.Disabled
 import com.automattic.android.tracks.crashlogging.PerformanceMonitoringConfig.Enabled
 import com.automattic.android.tracks.crashlogging.eventLevel
@@ -24,7 +26,6 @@ internal class SentryCrashLogging constructor(
     private val sentryWrapper: SentryErrorTrackerWrapper,
     applicationScope: CoroutineScope
 ) : CrashLogging {
-
     init {
         sentryWrapper.initialize(application) { options ->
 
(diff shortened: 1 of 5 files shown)


wzieba:
hi folks 👋 Just a heads-up that I recently bumped Sentry SDK to new major version #201 . Maybe we could sync this branch with `trunk` to make sure that the SDK bump didn't affect work here?

fluiddot:
> hi folks 👋 Just a heads-up that I recently bumped Sentry SDK to new major version #201 . Maybe we could sync this branch with `trunk` to make sure that the SDK bump didn't affect work here?

Hey @wzieba 👋 , thanks for the heads up 🙇 ! Sure, we'll update with `trunk` to confirm it works with the latest Sentry version.

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryCrashLogging.kt
```diff
@@ -134,6 +140,46 @@ internal class SentryCrashLogging constructor(
...
+            this.platform = "javascript"
+            this.appendTags(jsException.tags)
+            this.exceptions = mutableListOf(sentryException)
+        }
+
+        Sentry.configureScope { scope ->
+            scope.setContexts("react_native_context", jsException.context)
+        }
+        sentryWrapper.captureEvent(event)
+        callback.onReportSent(true)
```
fluiddot, line 180: @jhnstn I'm not sure if this will guarantee that the event has been sent at the time the callback is triggered. Did you have the chance to confirm it?

jhnstn: eeek, I'm not sure how to do that. I noticed you use some deferred queue and then call flush in the iOS tracker. 
I'm guessing we could use [coroutines](https://developer.android.com/kotlin/coroutines) here, but I'm not entirely sure why you use a deferred queue. Is the flush call blocking until the milliseconds run out? 

I did dig into the Sentry SDK to try to figure out what flush does, but I ran into this comment: 

```
Flushes events queued up to the current hub. Not implemented yet.
```


I can keep digging but I'm still unsure why we need this level of inspection. Do we expect `captureEvent` to fail often? can we catch the conditions where we think it will fail before calling?  Is there a concern that we might be silencing exceptions due to malformed expectation data?

fluiddot: > Is the flush call blocking until the milliseconds run out?

Yes. It tries to flush all queued Sentry events and finishes either when all events are sent or hitting the timeout ([reference](https://github.com/Automattic/Automattic-Tracks-iOS/blob/d3694e8a352f8cac3f3c26f73125614408c01abc/Sources/Remote%20Logging/Crash%20Logging/CrashLogging.swift#L244)). I haven't checked if the Android implementation works the same way, but if the events are being queued, we'd need to find a way to ensure they are sent. As I shared [here](https://github.com/WordPress/gutenberg/pull/59221#discussion_r1508682107), the main purpose of the callback is to ensure the event is sent to Sentry before letting the app crash. 

> Do we expect `captureEvent` to fail often? can we catch the conditions where we think it will fail before calling? Is there a concern that we might be silencing exceptions due to malformed expectation data?

We don't need to check for failures when sending the event. As I shared above, the callback's purpose is mainly to ensure that the event is sent.

@jhnstn This can be tested by forcing an unhandled exception and checking in Sentry if it gets reported before the app crashes. I'm currently testing this in installable builds.

fluiddot: I've just validated that the Sentry event is sent before the crash 🎊. Not sure if it works because on Android the crash only affects the editor screen (i.e. the app doesn't fully crash) or if Sentry events are not being queued. In any case, I confirm this approach works.

fluiddot (approved):
LGTM 🎊 !

PR #206, opened 2024-03-21
Allow consumers to delegate release name creation to the SDK

wzieba:
Main discussion: #204 

### Description

This PR introduces a possibility for SDK consumers to delegate defining a release name to the Tracks library (eventually - to Sentry SDK).

This is a way to address an issue of incorrectly assigning release names between projects in the same Sentry organization.

PR #209, opened 2024-03-27
fix: do not set `null` release if `SetByTracksLibrary` requested

wzieba:
### Description

This PR fixes my overlook introduced in #206 . Because `SentryOptions` initialization happens **after** Sentry SDK sets the default `release` name, if we set `null` in case of `SetByTracksLibrary`, we overwrite the value Sentry prepared.

I should've verified it before merge. Sorry for trouble.

### Fix
In the case of `SetByTracksLibrary` simply assign nothing related to releases

### Verification

Using a test project:

#### Before
image

#### After
`null` here is expected
image

### DIFF
--- a/AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryCrashLogging.kt
+++ b/AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryCrashLogging.kt
@@ -45,9 +45,8 @@ internal class SentryCrashLogging constructor(
             options.apply {
                 dsn = dataProvider.sentryDSN
                 environment = dataProvider.buildType
-                release = when (val releaseName = dataProvider.releaseName) {
-                    is ReleaseName.SetByApplication -> releaseName.name
-                    ReleaseName.SetByTracksLibrary -> null
+                (dataProvider.releaseName as? ReleaseName.SetByApplication)?.let {
+                    release = it.name
                 }
                 this.tracesSampleRate = tracesSampleRate
                 this.profilesSampleRate = profilesSampleRate


AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryCrashLogging.kt
```diff
@@ -45,9 +45,8 @@ internal class SentryCrashLogging constructor(
             options.apply {
                 dsn = dataProvider.sentryDSN
                 environment = dataProvider.buildType
-                release = when (val releaseName = dataProvider.releaseName) {
-                    is ReleaseName.SetByApplication -> releaseName.name
-                    ReleaseName.SetByTracksLibrary -> null
+                (dataProvider.releaseName as? ReleaseName.SetByApplication)?.let {
```
iangmaia, line 48: Hmm, couldn't you just

```kotlin
release = (dataProvider.releaseName as? ReleaseName.SetByApplication)?.name
```

wzieba: It's the same overlook I did in #206 😅. The version you provided **will** assign `null` to `release` if `dataProvider.releaseName` is not `SetByApplication`. And, in result, **will** overwrite the value Sentry SDK automatically assigned (what we wanted).

iangmaia: Aah gotcha, you don't really want the value to be set to `null` when the cast doesn't succeed 👍

iangmaia (approved):

AutomatticTracks/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryCrashLogging.kt
wzieba: Yes - when `dataProvider.releaseName` is `SetByTracksLibrary`, we should do nothing with `options.release`

iangmaia: > when dataProvider.releaseName is SetByTracksLibrary

You mean **isn't** ? 😅

wzieba: 🤔 `ReleaseName` can be either `SetByTracksLibrary` or `SetByApplication`.

When `dataProvider.releaseName` **is** `SetByTracksLibrary` we should do nothing with `options.release`. I think that's the correct description, matching the implementation.

iangmaia: Ah, I thought you were mentioning the same type that's in the code. All clear now, thanks!

PR #213, opened 2024-04-22
Introduce initialize method

### LINKED ISSUE #103: Provide `CrashLogging#initialize` method
### Description

Instead constructor-based initialisation, we should provide a separate `initialisation` method. Constructor-based one is tricky to use - if the consumer app wants to catch unhandled exceptions but doesn't want to send custom Sentry reports, it won't inject `CrashLogging` anywhere and therefore - won't make `SentryCrashLogging#init` to run.

This way, Sentry SDK won't be initialised and exceptions won't be catched.

Adding separate `initialisation` method will make SDK initialisation visible and transparent.

wzieba:
Closes #103 

## Description
This PR introduces `CrashLogging#initialize` method. The motivation for this is to reduce risks of issues like this one: https://github.com/Automattic/simplenote-android/pull/1518 . In other words: let's not assume that `CrashLogging` will be injected in the main application, but rather enforce it.

## Testing
You can edit `gradle.properties` and provide DSN of a test project (you can find it on Sentry) and report some events using sample app.


crashlogging/src/main/java/com/automattic/android/tracks/crashlogging/CrashLogging.kt
```diff
@@ -1,6 +1,12 @@
 package com.automattic.android.tracks.crashlogging
 
 interface CrashLogging {
+
+    /**
+     * Initializes the crash logging service with the provided by [CrashLoggingDataProvider] configuration.
```
iangmaia:
```suggestion
     * Initializes the crash logging service with the configuration provided by [CrashLoggingDataProvider].
```

crashlogging/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryCrashLogging.kt
```diff
@@ -27,12 +27,13 @@ import kotlinx.coroutines.CoroutineScope
...
-    application: Application,
+    private val application: Application,
     private val dataProvider: CrashLoggingDataProvider,
     private val sentryWrapper: SentryErrorTrackerWrapper,
-    applicationScope: CoroutineScope
+    private val applicationScope: CoroutineScope
 ) : CrashLogging {
-    init {
+
+    override fun initialize() {
```
iangmaia, line 39: I don't think this is a must (or if you think it makes sense), but perhaps it can be useful to throw some `IllegalStateException` or similar to indicate the lack of initialization when performing some key action(s).

crashlogging/src/main/java/com/automattic/android/tracks/crashlogging/internal/ApplicationInfoProvider.kt
```diff
@@ -0,0 +1,14 @@
...
+
+internal interface ApplicationInfoProvider {
+    val debuggable: Boolean
+}
+
+internal class AndroidApplicationInfoProvider(context: Context) : ApplicationInfoProvider {
+    override val debuggable: Boolean = context.applicationInfo.let {
+        it.flags and ApplicationInfo.FLAG_DEBUGGABLE != 0
+    }
+}
```
wzieba, line 14: It's not really possible to test without setting up Robolectric or instrumantation tests. I think it's too much effort for this in this library, but I've validated this behavior on Day One Android: the app crashes in debug but not in release if reporting without initialization:

Screenshot 2024-04-24 at 15 58 18

crashlogging/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryCrashLogging.kt
wzieba: Thanks @iangmaia . I was going back and forth with this, but motivated by your comment, I've added such check in https://github.com/Automattic/Automattic-Tracks-Android/pull/213/commits/fbcd738ee08dca818b1e90dc6626fdffc76dad83 . It works only in `debug` - I don't want to risk throwing unhandled exception in release for "uninitialized crash logger". But in `debug` it'll be helpful to spot any initialization issues.

wzieba:
Thanks @iangmaia , ready for the second round!

crashlogging/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryCrashLogging.kt
iangmaia: I think it's a nice compromise, as developers using the SDK will be able to course correct sooner if they're missing something.

Thanks for considering it!

iangmaia (approved):
:shipit:

PR #215, opened 2024-04-29
Remove attaching Tracks version to user agent

### LINKED ISSUE #157: Remove library version information from tracking events
Remove [`LIB_VERSION`](https://github.com/Automattic/Automattic-Tracks-Android/blob/trunk/AutomatticTracks/src/main/java/com/automattic/android/tracks/TracksClient.java#L33). More context internally: pdnsEh-IK-p2

wzieba:
Closes: #157 

### Description
It was outdated and has no real usage.

### DIFF
--- a/AutomatticTracks/build.gradle
+++ b/AutomatticTracks/build.gradle
@@ -30,7 +30,6 @@ android {
     defaultConfig {
         minSdkVersion rootProject.minSdkVersion
         targetSdkVersion rootProject.targetSdkVersion
-        buildConfigField "String", "VERSION_NAME", "\"2.1.0\""
     }
 
     buildFeatures {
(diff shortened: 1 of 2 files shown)


AutomatticTracks/build.gradle
```diff
@@ -30,7 +30,6 @@ android {
     defaultConfig {
         minSdkVersion rootProject.minSdkVersion
         targetSdkVersion rootProject.targetSdkVersion
-        buildConfigField "String", "VERSION_NAME", "\"2.1.0\""
```
iangmaia, line 33: Isn't it handy to know what's the version a client has for troubleshooting, though? Is there any other way of knowing that?

wzieba: Over few years this property wasn't used AFAIK. Also, if we'll spot a problem, we can always backtrack the Tracks version looking at the client.

Taking the correct version of the Tracks SDK is not straightforward - we assign version number during the build time based on git parameters. There's no API for that now, and it's not really planned. More context: pdnsEh-IK-p2#comment-1705

iangmaia: Got it! Thanks for the context.
Perhaps if eventually we have a much wider use of Tracks (to observe things like version adoption, complaints, etc) it would make more sense to have the version.

iangmaia (approved):

PR #217, opened 2024-05-02
Allow consumers to specify sample rate

wzieba:
### Description

This PR introduces a new property to configure for consumers: `errorSampling` which allows them to reduce the number of issues they report.

### Testing
It's not necessarily - the property is passed (unit tests) and we don't aim to test Sentry framework here - we trust it simply works.

### DIFF
--- a/crashlogging/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryCrashLogging.kt
+++ b/crashlogging/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryCrashLogging.kt
@@ -75,6 +75,7 @@ internal class SentryCrashLogging constructor(
                     appendExtra(event)
                     event
                 }
+                sampleRate = dataProvider.errorsSampleRate
             }
         }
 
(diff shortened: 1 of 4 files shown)


wzieba:
hi @iangmaia 👋 are you fine with reviewing this PR? 🙂

crashlogging/src/main/java/com/automattic/android/tracks/crashlogging/internal/SentryCrashLogging.kt
```diff
@@ -75,6 +76,10 @@ internal class SentryCrashLogging constructor(
                     appendExtra(event)
                     event
                 }
+                sampleRate = when (val errorsSampleRate = dataProvider.errorSampling) {
+                    ErrorSampling.Disabled -> null
```
iangmaia, line 80: The `null` here confused me a bit, but looking at the code it's clear that this is a sort of mapping between the exposed `ErrorSampling` and this internal `options` from Sentry, I guess?
No change required, I just want to confirm my understanding 🙂 (but maybe worth a comment if it's not that obvious)

wzieba: > The null here confused me a bit, but looking at the code it's clear that this is a sort of mapping between the exposed ErrorSampling and this internal options from Sentry, I guess?

Yes, exactly! Looking at source code of Sentry, `options.sampleRate` is `null` by default and `null` means "we don't sample errors". 

Alternatively, `null` is the same as we would set `sampleRate` = `1.0` (meaning: all errors are sent), but `null` is a default expected by framework to disable this feature, so that's why I decided to map like this.

iangmaia (approved):

PR #218, opened 2024-05-02
Make user details nullable and null by default

wzieba:
### Description

Some projects, e.g. Pocket Casts, doesn't have user id or username. We shouldn't force those values - empty strings will be just a noise

### DIFF
--- a/crashlogging/src/main/java/com/automattic/android/tracks/crashlogging/CrashLoggingUser.kt
+++ b/crashlogging/src/main/java/com/automattic/android/tracks/crashlogging/CrashLoggingUser.kt
@@ -1,7 +1,7 @@
 package com.automattic.android.tracks.crashlogging
 
 data class CrashLoggingUser(
-    val userID: String,
-    val email: String,
-    val username: String
+    val userID: String? = null,
+    val email: String? = null,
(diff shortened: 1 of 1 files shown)


crashlogging/src/main/java/com/automattic/android/tracks/crashlogging/CrashLoggingUser.kt
```diff
@@ -1,7 +1,7 @@
 package com.automattic.android.tracks.crashlogging
 
 data class CrashLoggingUser(
```
iangmaia, line 3: I'm guessing you can have only one or two but not all properties, right?
If this was "all or nothing", you could also make `CrashLoggingUser?`, but I assume some clients will have only part of the properties.

iangmaia (approved):

crashlogging/src/main/java/com/automattic/android/tracks/crashlogging/CrashLoggingUser.kt
wzieba: Yes, exactly. E.g. in case of Pocket Casts, users only have mails. I'd like to make this API flexible for this.

PR #219, opened 2024-05-08
Track SDK version code

MiSikora:
`Build.VERSION.RELEASE` is not very developer-friendly. It can contain things like `8.0.0`, `8.1.0`, `8.1Go`, `12.0`, and so on. It is not normalized and it is not super useful when trying to cross-reference things with the documentation.

This PR ads `SDK_INT` which corresponds to API level of Android installed on a device.


wzieba (approved):
Looks good, thanks!

PR #221, opened 2024-05-21
Move Sentry config to CrashLogging manifest

wzieba:
This PR moves Sentry's configuration declared in `AndroidManifest.xml` from `tracks` to `crashlogging`.

This way, if a project uses only `crashlogging` module, they will have the correct Sentry setup.

It's a follow-up to #212


ParaskP7 (approved):
👋 @wzieba !

I have reviewed this PR and everything works as expected, good job! 🌟 

-----

I also went on a bit further as to test this PR, using the sample app and my own sentry test project DSN, everything looks to be working as expected. ✅ 

PS: I noticed on #212 that you specified that testing is not necessary, but, it being a breaking change and all, I thought I better (briefly) test it anyway. 😊

PR #223, opened 2024-06-28
Fix all Simplenote users and Pocket Cast users being set to Day One users

MiSikora:
Switch on user type used fall-through for Simplenote and Pocket Casts users, which marked them as Day One ones.


wzieba (approved):
😅 thanks for the fix. LGTM

PR #235, opened 2024-08-20
[Dependency Analysis] Remove Unused Dependencies

ParaskP7:
Project Thread: paaHJt-6YV-p2

### Description

Based on the `unused` related advises generated by the [Dependency Analysis](https://github.com/autonomousapps/dependency-analysis-gradle-plugin) Gradle plugin, this PR removes all unused dependencies from this project.

FYI: As part of this change, the [Dependency Analysis](https://github.com/autonomousapps/dependency-analysis-gradle-plugin) Gradle plugin got also updated to its latest [1.33.0](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/main/CHANGELOG.md#version-1330) version.

### Testing Steps

1. Tooling:
    - Run the `./gradlew buildHealth` task and verify that there is no `unused` related advise remaining within both reports, JSON and txt included:
        - `build-health-report.json` -> Search for `unusedCount`
        - `build-health-report.txt` -> Search for `unused`
    - Check the manually triggered [scheduled build](https://buildkite.com/automattic/automattic-tracks-android/builds/317) and verify that everything works as expected with this newer version of the [Dependency Analysis](https://github.com/autonomousapps/dependency-analysis-gradle-plugin) Gradle plugin ([1.33.0](https://github.com/autonomousapps/dependency-analysis-gradle-plugin/blob/main/CHANGELOG.md#version-1330)).
3. Testing:
    - Smoke test the sample tracks app, try out all available screens and functionality. Verify everything is working as expected.
    - Also, if you want to be thorough about reviewing the changes, you could quickly smoke test, either the [JP/WPAndroid](https://github.com/wordpress-mobile/WordPress-Android) and/or [WCAndroid](https://github.com/woocommerce/woocommerce-android), with this version of `Tracks`, or via composite builds, and see if it works as expected.

### DIFF
--- a/benchmark/build.gradle.kts
+++ b/benchmark/build.gradle.kts
@@ -43,6 +43,11 @@ android {
             isDefault = true
         }
     }
+
+    lint {
+        warningsAsErrors = true
+        lintConfig = file("${project.rootDir}/config/lint/lint.xml")
+    }
 }
 
 dependencies {
--- a/benchmark/build.gradle.kts
+++ b/benchmark/build.gradle.kts
@@ -57,3 +57,12 @@ dependencies {
     // https://developer.android.com/studio/projects/android-library#Convert
     androidTestImplementation(project(":AutomatticTracks"))
 }
+
+dependencyAnalysis {
+    issues {
+        onUnusedDependencies {
+            // This dependency is not needed but kept to preserve the module's default configuration.
+            exclude("org.jetbrains.kotlin:kotlin-stdlib")
+        }
+    }
+}
--- a/benchmark/build.gradle.kts
+++ b/benchmark/build.gradle.kts
@@ -62,12 +62,3 @@ dependencies {
     // https://developer.android.com/studio/projects/android-library#Convert
     androidTestImplementation(project(":AutomatticTracks"))
 }
-
-dependencyAnalysis {
-    issues {
-        onUnusedDependencies {
-            // This dependency is not needed but kept to preserve the module's default configuration.
-            exclude("org.jetbrains.kotlin:kotlin-stdlib")
-        }
-    }
-}
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -18,7 +18,7 @@ steps:
     plugins: [$CI_TOOLKIT]
     command: |
       cp gradle.properties-example gradle.properties
-      ./gradlew lintRelease ciktlint
+      ./gradlew lint ciktlint
     artifact_paths:
       - "**/build/reports/lint-results.*"
       - "**/build/ktlint.xml"
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,7 @@ proguard/
 
 # private config file
 gradle.properties
+!/benchmark/gradle.properties
 
 # OS X generated file
 .DS_Store
--- a/AutomatticTracks/build.gradle
+++ b/AutomatticTracks/build.gradle
@@ -17,7 +17,6 @@ repositories {
 
 dependencies {
     implementation "androidx.annotation:annotation:$androidxAnnotationVersion"
-    implementation "com.squareup.okhttp3:okhttp:$squareupOkhttpVersion"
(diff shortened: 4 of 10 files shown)


benchmark/build.gradle.kts
```diff
@@ -57,3 +62,12 @@ dependencies {
...
 }
+
+dependencyAnalysis {
+    issues {
+        onUnusedDependencies {
+            // This dependency is not needed but kept to preserve the module's default configuration.
+            exclude("org.jetbrains.kotlin:kotlin-stdlib")
+        }
+    }
+}
```
wzieba: I have similar concern as in other, similar scenario. If `kotlin-stdlib` is not required, then let's maybe instruct kotlin gradle plugin to... not include `kotlin-stdlib`.

It won't affect build in any way, as this is about adding a dependency to project's classpath (not build classpath). The authors of the plugin also offered a ready solution for this: https://kotlinlang.org/docs/gradle-configure-project.html#dependency-on-the-standard-library so we don't hack anything.

benchmark/build.gradle.kts
```diff
@@ -43,6 +43,11 @@ android {
             isDefault = true
         }
     }
+
+    lint {
+        warningsAsErrors = true
+        lintConfig = file("${project.rootDir}/config/lint/lint.xml")
+    }
```
wzieba, line 50: I'm not sure if we should care about lint issues in a `benchmark` module, it's purely utilitarian one, used only to generate performance benchmarks when it's necessary.

wzieba (approved):

benchmark/build.gradle.kts
ParaskP7: 👋 @wzieba !

> ...then let's maybe instruct kotlin gradle plugin to... not include kotlin-stdlib.

To my understanding, you suggest adding the `kotlin.stdlib.default.dependency=false` within `gradle.properties`, but have you (maybe) tried that yourself? This, as I would also expect, fails the build, because although this `benchmark` module doesn't need the `kotlin-stdlib` dependency, all other modules do. Isn't that right? 🤔

What we want is to tell our build to ignore the `kotlin-stdlib` dependency, yes, but for `benchmark` only. I am not sure what's the correct way to do that, especially since this also needs to be done on the source-set level. 🤷 

PS: I think that for what is worth we shouldn't care too much about the configuration of this module, it is just an auto-generate benchmark module that serve a very specific purpose.

benchmark/build.gradle.kts
ParaskP7: This is true @wzieba , but then again, during development, while I was running `./gradlew lintDebug` to verify things, Lint kept failing for me. Actually, I kept hitting that problem for months now, but never bother to do anything about it. Now, and since the configuration was very straightforward, I wanted to fix so that we Lint never fails again for this project, no matter the module.

benchmark/build.gradle.kts
wzieba: Each module can have its own `gradle.properties`. Having `kotlin.stdlib.default.dependency=false` in the `gradle.properties` in `benchmark` works just fine

ParaskP7: True @wzieba , I just tried it and it indeed works, I am curious why though, as I would have expected that `./gradlew benchmark:assembleAndroidTest` to fail since `TrackingMethodsBenchmark` is written in Kotlin, but it works and the project compiles, strange... 🤔 

No matter, this is now done: a2a349dff6166c6d9fe56b200fa80a7ff02df7b6

wzieba: Thanks! Just sharing that the stdlib contains some additional features, but it's not necessary to compile Kotlin code - that's why this module still compiles.

ParaskP7: Ah true @wzieba (😅 ) , it is just a Kotlin standard library with some additional features... I didn't have to deal with that for so long that I am now forgetting things, thanks for the merge! 🤷

PR #236, opened 2024-08-20
Bump ktlint to the latest backward compatible version

wzieba:
Currently used version, `0.47.1` crashes in some cases and it slows down my development.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,7 +27,7 @@ subprojects {
     apply from: "$project.rootDir/ktlint.gradle"
 
     dependencies {
-        ktlint "com.pinterest:ktlint:0.47.1"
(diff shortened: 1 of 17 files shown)


wzieba:
Actually, the crash was not caused by `0.47.1` but by ae1abe5aed0b921905bd0ba3e9965d10ed43d502 🫣. But anyway, I think it's fine to update.

ParaskP7 (approved):
👋 @wzieba !

I have reviewed and also tested this PR (`./gradlew ktlint`), everything seems to be working as expected, good job! 🌟 

PS: I know that this is out-of-scope, but next time I would recommend to instead replace KtLint with Detekt ([related](https://github.com/orgs/Automattic/projects/841/views/6?pane=issue&itemId=71410072)).

PR #237, opened 2024-08-20
Introduce `ExperimentRestClient` and basic domain models

wzieba:
### Description

This PR is a first one of a few that aim to remove `FluxC` as a dependency for `experimantation` project.

In this PR I introduce `ExperimentRestClient` with logic of parsing fetched JSON and mapping it to domain models.

I decided to not introduce a feature branch to keep things simpler: newly introduced classes are not visible to consumers (except domain models, which are useless on its own) so it shouldn't cause confusion.

The logic is based on FluxC implementation, especially `org.wordpress.android.fluxc.network.rest.wpcom.experiments.ExperimentRestClient`

### How to test?

At this moment, testing is not required or possible due to lack of an entry point. The testing is covered with unit tests, and in next PRs I'll prepare some test scenarios.

### DIFF
--- a/experimentation/build.gradle
+++ b/experimentation/build.gradle
@@ -45,7 +45,7 @@ dependencies {
     ksp "com.squareup.moshi:moshi-kotlin-codegen:$moshi"
 
     // Coroutines
-    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:$kotlinxCoroutinesExperimentationVersion"
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:$kotlinxCoroutinesVersion"
 
     //FluxC
     implementation "org.wordpress:fluxc:$wordPressFluxCVersion"
@@ -53,13 +53,14 @@ dependencies {
     //WordPress utils
     implementation "org.wordpress:utils:$wordPressUtilsVersion"
 
-    testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:$kotlinxCoroutinesExperimentationVersion"
+    testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:$kotlinxCoroutinesVersion"
     testImplementation "junit:junit:$junitVersion"
     testImplementation "org.assertj:assertj-core:$assertjVersion"
     testImplementation "org.mockito.kotlin:mockito-kotlin:$mockitoKotlinVersion"
     testImplementation "org.mockito:mockito-inline:$mockitoInlineVersion"
 
     testImplementation "com.squareup.okhttp3:mockwebserver:$okHttp"
+    testImplementation "org.jetbrains.kotlin:kotlin-test:${gradle.ext.kotlinVersion}"
 }
 
 afterEvaluate {
--- a/experimentation/build.gradle
+++ b/experimentation/build.gradle
@@ -62,6 +62,15 @@ dependencies {
     testImplementation "com.squareup.okhttp3:mockwebserver:$okHttp"
 }
 
+dependencyAnalysis {
+    issues {
+        onUsedTransitiveDependencies {
+            // It doesn't seem necessary to directly declare this
+            exclude("com.squareup.okio:okio")
+        }
+    }
+}
+
 afterEvaluate {
     publishing {
         publications {
--- a/experimentation/build.gradle
+++ b/experimentation/build.gradle
@@ -2,7 +2,7 @@ plugins {
     id "com.android.library"
     id "org.jetbrains.kotlin.android"
     id 'com.automattic.android.publish-to-s3'
-    id "com.google.devtools.ksp" version "1.9.24-1.0.20"
+    id "com.google.devtools.ksp"
 }
 
 repositories {
--- a/experimentation/build.gradle
+++ b/experimentation/build.gradle
@@ -60,7 +60,6 @@ dependencies {
     testImplementation "org.mockito:mockito-inline:$mockitoInlineVersion"
 
     testImplementation "com.squareup.okhttp3:mockwebserver:$okHttp"
-    testImplementation "org.jetbrains.kotlin:kotlin-test:${gradle.ext.kotlinVersion}"
 }
 
 afterEvaluate {
--- a/experimentation/build.gradle
+++ b/experimentation/build.gradle
@@ -2,6 +2,7 @@ plugins {
     id "com.android.library"
     id "org.jetbrains.kotlin.android"
     id 'com.automattic.android.publish-to-s3'
+    id "com.google.devtools.ksp" version "1.9.24-1.0.20"
 }
 
 repositories {
@@ -36,6 +37,13 @@ android {
 }
 
 dependencies {
+    def okHttp = "4.12.0"
+    implementation "com.squareup.okhttp3:okhttp:$okHttp"
+
+    def moshi = "1.15.1"
+    implementation "com.squareup.moshi:moshi:$moshi"
+    ksp "com.squareup.moshi:moshi-kotlin-codegen:$moshi"
+
     // Coroutines
     implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:$kotlinxCoroutinesExperimentationVersion"
 
@@ -50,6 +58,8 @@ dependencies {
     testImplementation "org.assertj:assertj-core:$assertjVersion"
     testImplementation "org.mockito.kotlin:mockito-kotlin:$mockitoKotlinVersion"
     testImplementation "org.mockito:mockito-inline:$mockitoInlineVersion"
+
+    testImplementation "com.squareup.okhttp3:mockwebserver:$okHttp"
 }
 
 afterEvaluate {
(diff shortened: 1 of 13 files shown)


experimentation/build.gradle
```diff
@@ -36,20 +37,30 @@ android {
 }
 
 dependencies {
+    def okHttp = "4.12.0"
+    implementation "com.squareup.okhttp3:okhttp:$okHttp"
```
ParaskP7, line 41: Minor (🔍): FYI: There exist an advise from `Dependency Analysis` on transitive dependencies related to `okio`:

```
These transitive dependencies should be declared directly:
  implementation 'com.squareup.okio:okio:3.7.0'
```

experimentation/build.gradle
```diff
@@ -36,20 +37,30 @@ android {
...
 
-    testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:$kotlinxCoroutinesExperimentationVersion"
+    testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:$kotlinxCoroutinesVersion"
     testImplementation "junit:junit:$junitVersion"
     testImplementation "org.assertj:assertj-core:$assertjVersion"
     testImplementation "org.mockito.kotlin:mockito-kotlin:$mockitoKotlinVersion"
     testImplementation "org.mockito:mockito-inline:$mockitoInlineVersion"
+
+    testImplementation "com.squareup.okhttp3:mockwebserver:$okHttp"
+    testImplementation "org.jetbrains.kotlin:kotlin-test:${gradle.ext.kotlinVersion}"
```
ParaskP7: Minor (🔍): FYI: There exist an advise from `Dependency Analysis` on transitive dependencies related to `kotlin-test`:

```
Unused dependencies which should be removed:
  testImplementation('org.jetbrains.kotlin:kotlin-test:1.9.24') { capabilities {
    requireCapability('org.jetbrains.kotlin:kotlin-test-framework-junit')
}}
```

experimentation/src/main/java/com/automattic/android/experimentation/remote/AssignmentsDtoMapper.kt
```diff
@@ -0,0 +1,20 @@
+package com.automattic.android.experimentation.remote
+
+import com.automattic.android.experimentation.domain.Assignments
+import com.automattic.android.experimentation.domain.Variation
+
+internal object AssignmentsDtoMapper {
```
ParaskP7, line 6: Suggestion (💡): Consider adding a test for the mapper, just because of the `control` variation and it being mapped from `null` or `control`.

experimentation/src/main/java/com/automattic/android/experimentation/remote/AssignmentsDtoMapper.kt
```diff
@@ -0,0 +1,20 @@
...
+    fun AssignmentsDto.toAssignments(fetchedAt: Long): Assignments {
+        return Assignments(
+            variations = variations.mapValues { (key, value) ->
```
ParaskP7: Minor (🔍): Destructured parameter `key` is never used, consider renaming it to `_`.

experimentation/src/main/java/com/automattic/android/experimentation/remote/ExperimentRestClient.kt
```diff
@@ -0,0 +1,47 @@
...
+
+        val request = Request.Builder()
+            .url(url)
+            .get()
+            .build()
+
+        return withContext(Dispatchers.IO) {
+            okHttpClient.newCall(request).execute().use { response ->
+                if (!response.isSuccessful) {
+                    Result.failure<Assignments>(IOException("Unexpected code $response"))
```
ParaskP7: Minor (🔍): Consider remove the explicit `<Assignments>` type arguments.

experimentation/src/main/java/com/automattic/android/experimentation/remote/ExperimentRestClient.kt
```diff
@@ -0,0 +1,47 @@
...
+                } else {
+                    runCatching {
+                        val dto = jsonAdapter.fromJson(response.body!!.source())!!
```
ParaskP7, line 40: Question (❓): You prefer using `runCatching` with `!!` over explicitly returning a success/failure result, right? Is there a way for us to get rid of `!!` with `runCatching` or you don't mind as anything will flow through the default failure when that happens anyway? 🤔

experimentation/src/test/java/com/automattic/android/experimentation/remote/ExperimentRestClientTest.kt
```diff
@@ -0,0 +1,82 @@
...
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.runTest
+import okhttp3.mockwebserver.MockResponse
+import okhttp3.mockwebserver.MockWebServer
+import org.junit.Before
+import org.junit.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+internal class ExperimentRestClientTest {
```
ParaskP7, line 16: Praise (❤️): Good job introducing this "integration" test with `MockWebServer` here!

ParaskP7 (approved):
👋 @wzieba !

I have reviewed this PR as per the instructions, everything LGTM, good job! 🌟 

FYI: I also traversed the [FluxC](https://github.com/wordpress-mobile/WordPress-FluxC-Android) related repo and its [ExperimentRestClient](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/67b716889eb864763d703eb6d8c7519f73ad3157/fluxc/src/main/java/org/wordpress/android/fluxc/network/rest/wpcom/experiments/ExperimentRestClient.kt) + [AssignmentsModel](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/67b716889eb864763d703eb6d8c7519f73ad3157/fluxc/src/main/java/org/wordpress/android/fluxc/model/experiments/AssignmentsModel.kt) implementation and everything seems to be mapped 1-to-1 with your `Moshi` solution here, well done. 🥇 

> I decided to not introduce a feature branch to keep things simpler: newly introduced classes are not visible to consumers (except domain models, which are useless on its own) so it shouldn't cause confusion.

👍 

> At this moment, testing is not required or possible due to lack of an entry point. The testing is covered with unit tests, and in next PRs I'll prepare some test scenarios.

👍 

-----

I have left one question (❓), a suggestion (💡) and some minor (🔍) comments for you to consider. I am going to approve this PR anyway, since none is blocking. I am **NOT** going to merge this PR yet to give you some time to apply any of my suggestions. However, feel free to ignore them and merge the PR yourself.

experimentation/build.gradle
wzieba: I wonder why though - we don't use any code from `okio` directly 🤔

ParaskP7: Yea, I was wondering about that too, maybe the generate code by `Moshi` has to do with that, not sure, didn't look at the dependency graph, just searched for import and me too couldn't finding anything `Okio` related? 🤔 

FYI: We can suppress/exclude that if need be.

experimentation/build.gradle
wzieba: Nice, TIL about `capabilities` 😮

wzieba: Anyway, I removed this library because I used it only for assertions and we can use JUnit for this.

experimentation/src/main/java/com/automattic/android/experimentation/remote/ExperimentRestClient.kt
wzieba: I think `runCatching` is a good fit because it returns the `Result` type - just what we need here. It protects against the NPE (from `!!`) but also from an incorrect JSON we parse here.

> Is there a way for us to get rid of !! with runCatching or you don't mind as anything will flow through the default failure when that happens anyway? 🤔

I don't mind much here about it 😅 I think it's okay to wrap here all exceptions - I understand it's usually an anti-pattern, but we don't want ever to have uncaught exception in `fetchAssignments`

experimentation/src/main/java/com/automattic/android/experimentation/remote/AssignmentsDtoMapper.kt
wzieba: Done 👍 41b49391b3ce1fd87b06f4d1c7629b4056ad40c5

wzieba:
Thank you for the review @ParaskP7 ! I've applied your suggestions, see you in next PRs 🙌

experimentation/build.gradle
ParaskP7: Makes sense to me, thanks! 👍

experimentation/src/main/java/com/automattic/android/experimentation/remote/AssignmentsDtoMapper.kt
ParaskP7: Awesome, the test LGTM, thanks! ❤️

experimentation/src/main/java/com/automattic/android/experimentation/remote/ExperimentRestClient.kt
ParaskP7: Let's then move forward with this solution! 👍 

> ...but also from an incorrect JSON we parse here.

💯

PR #238, opened 2024-08-21
Introduce file-based persistence for Experimentation module

wzieba:
### Description

Similarly to #237 , this PR is part of "defluxcification" of experiments module.

In this PR, I'm introducing a persistence layer that will work as a cache. In original implementation (FluxC), this was handled via saving a json file in shared preferences. I think a simple file would fit here also well, I don't see a compiling arguments to use shared preferences, while it introduces some more complexity.

### How to test?

The same as in #237 at this moment unit tests have to suffice. In next PRs, I'll bring some test cases that will cover the persistance.

### DIFF
--- a/experimentation/src/main/java/com/automattic/android/experimentation/local/FileBasedCache.kt
+++ b/experimentation/src/main/java/com/automattic/android/experimentation/local/FileBasedCache.kt
@@ -1,0 +1,37 @@
+package com.automattic.android.experimentation.local
+
+import com.automattic.android.experimentation.domain.Assignments
+import com.automattic.android.experimentation.remote.AssignmentsDtoJsonAdapter
+import com.automattic.android.experimentation.remote.AssignmentsDtoMapper.toAssignments
+import com.automattic.android.experimentation.remote.AssignmentsDtoMapper.toDto
+import com.squareup.moshi.Moshi
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import java.io.File
+
+internal class FileBasedCache(
+    private val cacheDir: File,
+    private val moshi: Moshi = Moshi.Builder().build(),
+    private val jsonAdapter: AssignmentsDtoJsonAdapter = AssignmentsDtoJsonAdapter(moshi),
+) {
+
+    private val assignmentsFile = File(cacheDir, "assignments.json")
+
+    suspend fun getAssignments(): Assignments? {
+        return withContext(Dispatchers.IO) {
+            assignmentsFile.takeIf { it.exists() }?.readText()?.let { json ->
+                val dto = jsonAdapter.fromJson(json)
+                dto?.toAssignments(null)
+            }
+        }
+    }
+
+    suspend fun saveAssignments(assignments: Assignments) {
+        withContext(Dispatchers.IO) {
+            val dto = assignments.toDto()
+            val json = jsonAdapter.serializeNulls().toJson(dto)
+
+            assignmentsFile.writeText(json)
+        }
+    }
+}
--- a/experimentation/src/main/java/com/automattic/android/experimentation/local/FileBasedCache.kt
+++ b/experimentation/src/main/java/com/automattic/android/experimentation/local/FileBasedCache.kt
@@ -5,6 +5,7 @@ import com.automattic.android.experimentation.remote.AssignmentsDtoJsonAdapter
 import com.automattic.android.experimentation.remote.AssignmentsDtoMapper.toAssignments
 import com.automattic.android.experimentation.remote.AssignmentsDtoMapper.toDto
 import com.squareup.moshi.Moshi
+import com.squareup.moshi.Types
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
 import java.io.File
@@ -16,22 +17,34 @@ internal class FileBasedCache(
 ) {
 
     private val assignmentsFile = File(cacheDir, "assignments.json")
+    private val type = Types.newParameterizedType(
+        Map::class.java,
+        Long::class.javaObjectType,
+        String::class.java,
+    )
+    private val wrapperAdapter = moshi.adapter<Map<Long, String>>(type)
 
     suspend fun getAssignments(): Assignments? {
         return withContext(Dispatchers.IO) {
             assignmentsFile.takeIf { it.exists() }?.readText()?.let { json ->
-                val dto = jsonAdapter.fromJson(json)
-                dto?.toAssignments(null)
+                val fromJson = wrapperAdapter.fromJson(json) ?: return@let null
+
+                val (fetchedAt, dtoJson) = fromJson.toList().first()
+
+                val dto = jsonAdapter.fromJson(dtoJson)
+
+                dto?.toAssignments(fetchedAt)
             }
         }
     }
 
     suspend fun saveAssignments(assignments: Assignments) {
         withContext(Dispatchers.IO) {
-            val dto = assignments.toDto()
-            val json = jsonAdapter.serializeNulls().toJson(dto)
+            val (dto, fetchedAt) = assignments.toDto()
+            val dtoJson = jsonAdapter.serializeNulls().toJson(dto)
 
-            assignmentsFile.writeText(json)
+            val wrapperJson = wrapperAdapter.toJson(mapOf(fetchedAt to dtoJson))
+            assignmentsFile.writeText(wrapperJson)
         }
     }
 }
--- a/experimentation/src/main/java/com/automattic/android/experimentation/remote/AssignmentsDtoMapper.kt
+++ b/experimentation/src/main/java/com/automattic/android/experimentation/remote/AssignmentsDtoMapper.kt
@@ -4,7 +4,7 @@ import com.automattic.android.experimentation.domain.Assignments
 import com.automattic.android.experimentation.domain.Variation
 
 internal object AssignmentsDtoMapper {
-    fun AssignmentsDto.toAssignments(externalFetchedAt: Long?): Assignments {
+    fun AssignmentsDto.toAssignments(fetchedAt: Long): Assignments {
         return Assignments(
             variations = variations.mapValues { (key, value) ->
                 if (value == null || value == "control") {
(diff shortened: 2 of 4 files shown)


experimentation/src/main/java/com/automattic/android/experimentation/local/FileBasedCache.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    suspend fun saveAssignments(assignments: Assignments) {
+        withContext(Dispatchers.IO) {
+            val (dto, fetchedAt) = assignments.toDto()
+            val dtoJson = jsonAdapter.serializeNulls().toJson(dto)
+
+            val wrapperJson = wrapperAdapter.toJson(mapOf(fetchedAt to dtoJson))
+            assignmentsFile.writeText(wrapperJson)
+        }
+    }
```
ParaskP7, line 49: Question (❓): On the [FluxC](https://github.com/wordpress-mobile/WordPress-FluxC-Android) side, there is this [clearCachedAssignments()](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/3f240ecc9161efca1d75183a29640bd21db95cb6/fluxc/src/main/java/org/wordpress/android/fluxc/store/ExperimentStore.kt#L54-L56) implementation as well, is that not needed for our case here?

experimentation/src/main/java/com/automattic/android/experimentation/local/FileBasedCache.kt
```diff
@@ -0,0 +1,50 @@
...
+                val (fetchedAt, dtoJson) = fromJson.toList().first()
+
+                val dto = jsonAdapter.fromJson(dtoJson)
+
+                dto?.toAssignments(fetchedAt)
+            }
+        }
+    }
+
+    suspend fun saveAssignments(assignments: Assignments) {
```
ParaskP7, line 41: Question (❓): On the [FluxC](https://github.com/wordpress-mobile/WordPress-FluxC-Android) side, the saving implementation is based on `AssignmentsModel` and not `Assignments` ([storeFetchedAssignments(...)](https://github.com/wordpress-mobile/WordPress-FluxC-Android/blob/3f240ecc9161efca1d75183a29640bd21db95cb6/fluxc/src/main/java/org/wordpress/android/fluxc/store/ExperimentStore.kt#L43C17-L46)), which means that there is no need to have this `dto` + `fetchedAt` pair logic, nor lose the `control` information when mapping the `Variation.Control -> null`, as this `Variation.Control` can either mean `null` or `control`. Wdyt? 🤔

experimentation/src/main/java/com/automattic/android/experimentation/remote/AssignmentsDtoMapper.kt
```diff
@@ -17,4 +17,15 @@ internal object AssignmentsDtoMapper {
             fetchedAt = fetchedAt,
         )
     }
+    fun Assignments.toDto(): Pair<AssignmentsDto, Long> {
+        return AssignmentsDto(
+            variations = variations.mapValues { (_, value) ->
+                when (value) {
+                    is Variation.Control -> null
```
ParaskP7: Question (❓): Similar to the previous question, why did you chose to map this to `null` and not `control`? I understand that the information is lost anyway as `Variation.Control` can mean either one, but I think using `control` explicitly, even in cases of `null` feels more appropriate to me. Wdyt?

PS: Having said that, I personally think that saving the `AssignmentsDto` instead of `Assignments` might be the best, to avoid boilerplate, or losing information (no matter its importance) and match the [FluxC](https://github.com/wordpress-mobile/WordPress-FluxC-Android) implementation, just to be on the safest side.

experimentation/src/test/java/com/automattic/android/experimentation/local/FileBasedCacheTest.kt
```diff
@@ -0,0 +1,68 @@
...
+import kotlinx.coroutines.test.runTest
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNull
+import org.junit.Test
+import kotlin.io.path.createTempDirectory
+
+@OptIn(ExperimentalCoroutinesApi::class)
+internal class FileBasedCacheTest {
+
+    private lateinit var sut: FileBasedCache
```
ParaskP7: Minor (🔍): Consider deleting this `lateinit` field and instead directly using a `val` variable per test. I think doing that will make it much more straightforward that this is exactly what each test needs and removing this extra `field` related boilerplate.

- From: `sut = fileBasedCache()`
- To: `val sut = fileBasedCache()`

ParaskP7 (approved):
👋 @wzieba !

I have reviewed this PR as per the instructions, everything LGTM, good job! 🌟 

> In this PR, I'm introducing a persistence layer that will work as a cache. In original implementation (FluxC), this was handled via saving a json file in shared preferences. I think a simple file would fit here also well, I don't see a compiling arguments to use shared preferences, while it introduces some more complexity.

I think I agree with using a `cacheDir` instead of `SharedPreferences`, although I don't understand too much how `SharedPreferences` introduces more complexity, I think both have their own set of complexities.

PS: I am not sure if you've also considered using [DataStore](https://developer.android.com/topic/libraries/architecture/datastore) instead.

-----

I have left few questions (❓), and one minor (🔍) comment for you to consider. I am going to approve this PR anyway, since none is blocking. I am **NOT** going to merge this PR yet to give you some time to apply any of my suggestions. However, feel free to ignore them and merge the PR yourself.

experimentation/src/main/java/com/automattic/android/experimentation/local/FileBasedCache.kt
wzieba: My reasoning here is that remote layer shouldn't communicate directly with persistence. 

This rule is already broken in 98b9674f84f5e66acd8b1db14a9fe3eeac104e71 by not introducing a persistence-only model, but I did it consciously to not introduce a new, twin one.

I'd like to keep the unidirectional control flow: remote -> domain -> persistence though, hence using `Assigmnets` domain model here.

experimentation/src/main/java/com/automattic/android/experimentation/local/FileBasedCache.kt
wzieba: Good catch, thanks! I added this with my next work, which I have locally. I think it's not worth to cherry-pick it here, it'll land with other changes

experimentation/src/main/java/com/automattic/android/experimentation/local/FileBasedCache.kt
ParaskP7: Sold, I then agree with your reasoning of using `Assignments` domain model, even if makes it a bit more complicated, thanks for the clarification! 💯

experimentation/src/main/java/com/automattic/android/experimentation/remote/AssignmentsDtoMapper.kt
wzieba: > even in cases of null feels more appropriate to me. Wdyt?

I don't know - I think I don't share similar feeling, but I can apply such change if it makes things more readable.

Knowing we shouldn't base the internal-persistence on API, just sharing that API uses `null` to annotate control group, at least in these experiments I was checking.

experimentation/src/main/java/com/automattic/android/experimentation/local/FileBasedCache.kt
ParaskP7: Awesome, thanks! 👍

experimentation/src/test/java/com/automattic/android/experimentation/local/FileBasedCacheTest.kt
wzieba: Good point, thanks!

experimentation/src/main/java/com/automattic/android/experimentation/remote/AssignmentsDtoMapper.kt
wzieba: Addressed in 0c1b7bf

experimentation/src/test/java/com/automattic/android/experimentation/local/FileBasedCacheTest.kt
wzieba: Addressed in 4f8927f

experimentation/src/main/java/com/automattic/android/experimentation/remote/AssignmentsDtoMapper.kt
ParaskP7: I was thinking that the API is using `null` on rare cases, and instead is usually using `control`, but if that's not the case, then maybe using `null` is actually better, even if that doesn't make this more readable, not sure... After we've discussed this, I'm living this decision up to you. 👍 

PS: Btw, can we write a quite comment on `null` vs. `control` somewhere so that we at least are aware of the differences, usages or each, maybe this will help us in the future, or help other devs like us. Wdyt? 🤔

wzieba:
> PS: I am not sure if you've also considered using [DataStore](https://developer.android.com/topic/libraries/architecture/datastore) instead.

Yes, but then I thought it's even more complex than any of file or shared preferences-based solutions 😅 . It's neat for a few reasons, but for such simple case as we have here now, I decided to not introduce it.

experimentation/src/main/java/com/automattic/android/experimentation/remote/AssignmentsDtoMapper.kt
ParaskP7: > Addressed in https://github.com/Automattic/Automattic-Tracks-Android/commit/0c1b7bf0db499381f8bde0724450bad274702f4c

Good call on using a constant, actually, we should have used a constant from the start! 🥇

ParaskP7:
> Yes, but then I thought it's even more complex...

This is sad @wzieba but true nevertheless... 😅 🤷

experimentation/src/main/java/com/automattic/android/experimentation/remote/AssignmentsDtoMapper.kt
wzieba: Sure, added comment as well!

ParaskP7: Noted, thanks for that, I think we can merge this now! 🚀

PR #239, opened 2024-08-26
Experimentation: replace FluxC with internal components

wzieba:
### Description 

This PR replaces usage of `ExperimentsStore` from FluxC, with internal implementation of `AssignmentsRepository` that fetches data from remote, saves it to cache and offer it later.

I aimed for a minimal set of changes here, to not shadow that the logic remains unchanged here.

### How to test

Unit tests are passing after the change. But as they're heavily mocked, we can't really be sure that this caught all possible regressions: https://github.com/Automattic/Automattic-Tracks-Android/pull/239#discussion_r1732906646

Still, this is some safety net. To test this manually, please see #241

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -40,7 +40,6 @@ ext {
 
 ext {
     // libs
-    wordPressFluxCVersion = '1.36.0'
     wordPressUtilsVersion = '2.3.0'
 
     // main
--- a/build.gradle
+++ b/build.gradle
@@ -39,9 +39,6 @@ ext {
 
 
 ext {
-    // libs
-    wordPressUtilsVersion = '2.3.0'
-
     // main
     androidxAnnotationVersion = "1.1.0"
     androidxAppcompatVersion = '1.2.0'
--- a/experimentation/build.gradle
+++ b/experimentation/build.gradle
@@ -46,9 +46,6 @@ dependencies {
     // Coroutines
     implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:$kotlinxCoroutinesVersion"
 
-    //WordPress utils
-    implementation "org.wordpress:utils:$wordPressUtilsVersion"
-
     testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:$kotlinxCoroutinesVersion"
     testImplementation "junit:junit:$junitVersion"
     testImplementation "org.assertj:assertj-core:$assertjVersion"
--- a/experimentation/build.gradle
+++ b/experimentation/build.gradle
@@ -11,7 +11,6 @@ repositories {
         url "https://a8c-libs.s3.amazonaws.com/android"
         content {
             includeGroup "org.wordpress"
-            includeGroup "org.wordpress.fluxc"
             includeGroup "org.wordpress.wellsql"
         }
     }
@@ -46,9 +45,6 @@ dependencies {
 
     // Coroutines
     implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:$kotlinxCoroutinesVersion"
-
-    //FluxC
-    implementation "org.wordpress:fluxc:$wordPressFluxCVersion"
 
     //WordPress utils
     implementation "org.wordpress:utils:$wordPressUtilsVersion"
--- a/experimentation/build.gradle
+++ b/experimentation/build.gradle
@@ -11,7 +11,6 @@ repositories {
         url "https://a8c-libs.s3.amazonaws.com/android"
         content {
             includeGroup "org.wordpress"
-            includeGroup "org.wordpress.wellsql"
         }
     }
 }
--- a/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
+++ b/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
@@ -12,10 +12,9 @@ import com.automattic.android.experimentation.remote.ExperimentRestClient
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
-import org.wordpress.android.fluxc.store.ExperimentStore.Platform
 
 class ExPlat internal constructor(
-    private val platform: Platform,
+    private val platform: String,
     private val experiments: Set<Experiment>,
     private val logger: ExperimentLogger,
     private val coroutineScope: CoroutineScope,
@@ -98,7 +97,7 @@ class ExPlat internal constructor(
     }
 
     private suspend fun fetchAssignments() =
-        restClient.fetchAssignments(platform.value, experimentIdentifiers).fold(
+        restClient.fetchAssignments(platform, experimentIdentifiers).fold(
             onSuccess = {
                 logger.d("ExPlat: fetching assignments successful with result: $it")
             },
--- a/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
+++ b/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
@@ -3,23 +3,28 @@ package com.automattic.android.experimentation
 import com.automattic.android.experimentation.ExPlat.RefreshStrategy.ALWAYS
 import com.automattic.android.experimentation.ExPlat.RefreshStrategy.IF_STALE
 import com.automattic.android.experimentation.ExPlat.RefreshStrategy.NEVER
+import com.automattic.android.experimentation.domain.Assignments
+import com.automattic.android.experimentation.domain.AssignmentsValidator
+import com.automattic.android.experimentation.domain.Variation
+import com.automattic.android.experimentation.domain.Variation.*
+import com.automattic.android.experimentation.local.FileBasedCache
+import com.automattic.android.experimentation.remote.ExperimentRestClient
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
-import org.wordpress.android.fluxc.model.experiments.Assignments
-import org.wordpress.android.fluxc.model.experiments.Variation
-import org.wordpress.android.fluxc.model.experiments.Variation.Control
-import org.wordpress.android.fluxc.store.ExperimentStore
+import kotlinx.coroutines.runBlocking
 import org.wordpress.android.fluxc.store.ExperimentStore.Platform
 import org.wordpress.android.fluxc.utils.AppLogWrapper
 import org.wordpress.android.util.AppLog.T
 
-class ExPlat(
+class ExPlat internal constructor(
     private val platform: Platform,
     private val experiments: Set<Experiment>,
-    private val experimentStore: ExperimentStore,
     private val appLogWrapper: AppLogWrapper,
     private val coroutineScope: CoroutineScope,
     private val isDebug: Boolean,
+    private val cache: FileBasedCache,
+    private val assignmentsValidator: AssignmentsValidator,
+    private val restClient: ExperimentRestClient,
 ) {
     private val activeVariations = mutableMapOf<String, Variation>()
     private val experimentIdentifiers: List<String> = experiments.map { it.identifier }
@@ -49,7 +54,7 @@ class ExPlat(
         }
         return activeVariations.getOrPut(experimentIdentifier) {
             getAssignments(if (shouldRefreshIfStale) IF_STALE else NEVER)
-                .getVariationForExperiment(experimentIdentifier)
+                .variations[experimentIdentifier] ?: Control
         }
     }
 
@@ -64,7 +69,9 @@ class ExPlat(
     fun clear() {
         appLogWrapper.d(T.API, "ExPlat: clearing cached assignments and active variations")
         activeVariations.clear()
-        experimentStore.clearCachedAssignments()
+        coroutineScope.launch {
+            cache.clear()
+        }
     }
 
     private fun refresh(refreshStrategy: RefreshStrategy) {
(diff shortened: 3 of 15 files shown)


experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
```diff
@@ -74,20 +81,28 @@ class ExPlat(
     }
 
     private fun getAssignments(refreshStrategy: RefreshStrategy): Assignments {
-        val cachedAssignments = experimentStore.getCachedAssignments() ?: Assignments()
-        if (refreshStrategy == ALWAYS || (refreshStrategy == IF_STALE && cachedAssignments.isStale())) {
+        val cachedAssignments: Assignments =
+            runBlocking { repository.getCached() } ?: Assignments(emptyMap(), 0, 0)
```
wzieba, line 83: Re: `runBlocking`

We run file-based operation on the caller thread (risking it's a main thread). It was happening before with `experimentStore#getCachedAssignments`, but now it's just more visible.

I plan to address this in a future PR.

experimentation/src/test/java/com/automattic/android/experimentation/ExPlatTest.kt
```diff
@@ -83,14 +84,28 @@ class ExPlatTest {
 
         exPlat.forceRefresh()
 
-        verify(experimentStore, times(1)).fetchAssignments(eq(platform), any(), anyOrNull())
+        verify(restClient, times(1)).fetchAssignments(eq(platform), any(), anyOrNull())
+    }
+
+    @Test
+    fun `force refresh is successful when cache is fresh `() = runBlockingTest {
```
wzieba, line 91: I identify some issues with this tests suite. It's tightly coupled with implementation details of `ExPlat` and favors checks on method execution (all checks are `verify`) rather than actual outcomes.

For this reason, it doesn't really test things. E.g. the whole test suite was passing, even though I forgot to add logic for saving cache. See: 82aa328112475c28471bae45a3e0d4d76fda9558 (passing) but it has no saving logic.

The issue is that, e.g. we have "forceRefresh fetches assignments if cache is fresh", which checks if a `fetch` method was executed on a mock. We don't have verifications that assignments has been indeed really fetched. The, added by me in this PR, test "force refresh is successful when cache is fresh" is also not ideal: it checks if we called method on `cache`, but not that cache has been saved.

----

For all of these reasons I plan to work on improving this test suite, but I want to do this in a separate PR to not shadow changes here.

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
ParaskP7: > I plan to address this in a future PR.

Thanks for taking a note on that and trying to make this more "thread" robust! 🙏

build.gradle
```diff
@@ -39,10 +39,6 @@ ext {
 
 
 ext {
-    // libs
-    wordPressFluxCVersion = '1.36.0'
-    wordPressUtilsVersion = '2.3.0'
```
ParaskP7, line 44: Praise (❤️): This is awesome, with one stone you got two birds, `FluxC` and `Utils`, bullseye, this project is now lips dependency free!!! 🎉

experimentation/build.gradle
```diff
@@ -11,7 +11,6 @@ repositories {
         url "https://a8c-libs.s3.amazonaws.com/android"
         content {
             includeGroup "org.wordpress"
-            includeGroup "org.wordpress.fluxc"
             includeGroup "org.wordpress.wellsql"
```
ParaskP7: Suggestion (💡): Let's remove `wellsql` as well! 🎉

experimentation/src/main/java/com/automattic/android/experimentation/remote/AssignmentsDtoMapper.kt
```diff
@@ -16,7 +16,7 @@ internal object AssignmentsDtoMapper {
                     Variation.Treatment(value)
```
ParaskP7, line 16: Minor (🔍): After seeing `FluxC` being removed everywhere, I search for it one more time and we get a final reference to it on this comment above, consider updating the comment so that it won't mislead someone thinking that this has anything to do with `FluxC`.

experimentation/src/main/java/com/automattic/android/experimentation/repository/AssignmentsRepository.kt
```diff
@@ -0,0 +1,38 @@
...
+    ): Result<Assignments> {
+        val fetchResult =
+            experimentRestClient.fetchAssignments(platform, experimentNames, anonymousId)
+
+        return fetchResult.fold(
+            onFailure = {
+                Result.failure(it)
+            },
+            onSuccess = { assignments ->
+                cache.saveAssignments(assignments)
```
ParaskP7, line 25: Praise (❤️): Nice save (with the tests and [all](https://github.com/Automattic/Automattic-Tracks-Android/pull/239#discussion_r1732906646))!

experimentation/src/main/java/com/automattic/android/experimentation/repository/AssignmentsRepository.kt
```diff
@@ -0,0 +1,38 @@
...
+                Result.success(assignments)
+            },
+        )
+    }
+
+    suspend fun getCached(): Assignments? {
+        return cache.getAssignments()
+    }
+
+    suspend fun clear() {
```
ParaskP7: Minor (🔍): Consider renaming this to `clearCache()` or else maybe rename the above to removed the `Cached` suffix. 🤔

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
```diff
@@ -62,9 +67,11 @@ class ExPlat(
     }
 
     fun clear() {
-        appLogWrapper.d(T.API, "ExPlat: clearing cached assignments and active variations")
+        logger.d("ExPlat: clearing cached assignments and active variations")
         activeVariations.clear()
-        experimentStore.clearCachedAssignments()
+        coroutineScope.launch {
```
ParaskP7: Minor (🔍): Consider making this a one-liner like you did with the rest (including the `runBlocking` block).

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
```diff
@@ -74,20 +81,28 @@ class ExPlat(
     }
 
     private fun getAssignments(refreshStrategy: RefreshStrategy): Assignments {
-        val cachedAssignments = experimentStore.getCachedAssignments() ?: Assignments()
-        if (refreshStrategy == ALWAYS || (refreshStrategy == IF_STALE && cachedAssignments.isStale())) {
+        val cachedAssignments: Assignments =
+            runBlocking { repository.getCached() } ?: Assignments(emptyMap(), 0, 0)
+        if (refreshStrategy == ALWAYS || (
+                refreshStrategy == IF_STALE && assignmentsValidator.isStale(
```
ParaskP7: Minor (🔍): Consider making this a one-liner, IMHO this reads a bit better: `refreshStrategy == IF_STALE && assignmentsValidator.isStale(cachedAssignments))`

experimentation/src/test/java/com/automattic/android/experimentation/ExPlatTest.kt
```diff
@@ -15,21 +24,13 @@ import org.mockito.kotlin.times
 import org.mockito.kotlin.verify
 import org.mockito.kotlin.verifyNoInteractions
 import org.mockito.kotlin.whenever
-import org.wordpress.android.fluxc.model.experiments.Assignments
```
ParaskP7, line 18: Suggestion (💡): Consider replacing `runBlockingTest` with `runTest` everywhere.

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
```diff
@@ -74,20 +81,28 @@ class ExPlat(
     }
 
     private fun getAssignments(refreshStrategy: RefreshStrategy): Assignments {
```
ParaskP7, line 81: Question (❓): Please help me understand a bit more about this implementation here and how it actually being/should be utilized by any client of this library. 🙏 

1. To my understanding this function returns the (previously) cached assignments no matter what, then why is it fetching the assignment if refresh strategy is the right one or the assignment are stale? I really struggle to understand the logic here. I know you didn't write this, and only want to change that `FluxC` implementation 1:1 to what a `non-FluxC` should look like, but still, how does this `ExPlat` work in general. What I was expecting during that call is for the fetched assignments to be returned instead of the cached ones when the strategy/stale conditions are met. 🤔 
2. I then tried to research that myself real quick, searching the `ExPlat` keyword on both [JP/WPAndroid](https://github.com/wordpress-mobile/WordPress-Android) and [WCAndroid](https://github.com/woocommerce/woocommerce-android) and figured out that:
    1. [JP/WPAndroid](https://github.com/wordpress-mobile/WordPress-Android): This client is actually using its own version of `ExPlat`, right! 🤷 But, it shouldn't, should it? 😕 Maybe we need to update this client when we are ready with the `Tracks` new implementation, wdyt?  🤔
    2. [WCAndroid](https://github.com/woocommerce/woocommerce-android): This client is using the `Tracks` version of `ExPlat` but actually not utilizing it at all, right! 🤷 Is it actually needed there?
    3. To my understanding, no other client is using the `Tracks` version of `ExPlat`, just because of `FluxC`.

ParaskP7 (commented):
👋 @wzieba !

I have reviewed this PR as per the instructions, everything LGTM as expected, great job on finally removing `FluxC` as a dependency from `Tracks`, including removing `Utils` too! 🎉 x 🎉 ^ 🎉 

-----

I have left one questions (❓), a couple of suggestions (💡) and some minor (🔍) comments for you to consider. I am **NOT** going to approve this PR for now, nor block it, just comment on it. I do that because I want to first review all the other chained PRs first. But, if that breaks your planning, let me know and we can move forward accordingly.

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
wzieba: Good questions! I had similar confusions and pivots in understanding this.

Trying to answer broadly: we want `ExPlat` to not return different values for the same application session. So we initially return cached value and, under some circumstances, we request an update of the cache.

The update will be used only on the next application session or after app calls `ExPlat#clear` method (if e.g. user logs out).

**Why we don't want to return different values for the same application session?**

We want to protect SDK clients from a mistake, in which the would e.g. run half of the app processes in a `Control` group, then have an update of cache, and the rest of the app in `Treatment` group. This would lead to unexpected behaviors and failed experiments.

**Why we don't just fetch newest experiment assignment?**

We want this SDK to be async and not block clients with a `suspend` calls as some experiment changes might require e.g. different app initialization.

---

I hope this clarifies things. For the complete summary and explanations of the final design, please see #243 - I think everything might be more readable after the lecture.

---

> I then tried to research that myself real quick, searching the ExPlat keyword on both [JP/WPAndroid](https://github.com/wordpress-mobile/WordPress-Android) and [WCAndroid](https://github.com/woocommerce/woocommerce-android)

Yes, to my best knowledge, no Android mobile project at Automattic is doing A/B tests using ExPlat at this moment

ParaskP7: Awesome, thank you so much for clarifying my question(s) @wzieba , much much appriciated. Also, thanks so much for #243 and the `README.md` you are adding there, haven't got that far with my reviews yet, but I am sure this will further clarify any doubts I have about this library, it's public APIs, including how to use it.

> Yes, to my best knowledge, no Android mobile project at Automattic is doing A/B tests using ExPlat at this moment

This is a bit weird to me, if not risky... I wish we knew that this library is working as expected and that we have thoroughly tested it with some clients before having other clients using it, assuming it just works... 🤔

I know, maybe that library worked at some point on [WCAndroid](https://github.com/woocommerce/woocommerce-android), they are just not using it anymore, not having an experiment worth of this utility, but still, as time passes-by, having a stale unmaintained/untested library might create some weird issues we might not/cannot be aware of atm.

You addition a new set of buttons/fields (an experimentation screen) as part of #241 mitigates this risk for us, as we at least can verify that we see it working, which gives us confidence, but I want to raise that thought of mine anyway, that we should be careful and warn product teams about this library and it being "untested" to its fullest. Am I being paranoid here? 😅

wzieba: Not at all! I think we should request a safety check from product teams on this and, for now, release this as a `"beta"`.

ParaskP7: Cool, totally agree with your plan! 🎯

experimentation/src/test/java/com/automattic/android/experimentation/ExPlatTest.kt
wzieba: Good suggestion! I applied it in #242 . I'll skip doing this here, to not introduce merge conflicts.

experimentation/src/main/java/com/automattic/android/experimentation/remote/AssignmentsDtoMapper.kt
wzieba: I've adjusted the comment in 513eb17 . I left the `FluxC` name though, as I think "previous implementation" without any context is enigmatic and more difficult to look for. I hope this looks better now!

experimentation/src/test/java/com/automattic/android/experimentation/ExPlatTest.kt
ParaskP7: Awesome, thanks, will take a look at it there then. 👍

experimentation/src/main/java/com/automattic/android/experimentation/remote/AssignmentsDtoMapper.kt
ParaskP7: Makes sense, thanks for updating the comment @wzieba ! 🙇

PR #240, opened 2024-08-27
Support passing `anon_id` to ExPlat

wzieba:
Preferably, review:
- #239 

first.

### Description

Passing `anon_id` query parameter seems like a necessity for a part (all?) of the experiments.

I keep this nullable and not mandatory to align to iOS counterpart 

https://github.com/Automattic/Automattic-Tracks-iOS/blob/ccf6b7d7424e32857819d35aa54813875557f5c8/Sources/Experiments/ExPlatService.swift#L11

### Testing

Automated tests are introduced that reflect necessity of `anon_id` for some experiments. Further e2e manual testing is available with example app in #241

### DIFF
--- a/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
+++ b/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
@@ -24,11 +24,11 @@ class ExPlat internal constructor(
     private val activeVariations = mutableMapOf<String, Variation>()
     private val experimentIdentifiers: List<String> = experiments.map { it.identifier }
 
-    private var anonId: String? = null
+    private var anonymousId: String? = null
 
-    fun configure(anonId: String? = null) {
+    fun configure(anonymousId: String? = null) {
         clear()
-        this.anonId = anonId
+        this.anonymousId = anonymousId
     }
 
     /**
@@ -76,7 +76,7 @@ class ExPlat internal constructor(
     fun clear() {
         logger.d("ExPlat: clearing cached assignments and active variations")
         activeVariations.clear()
-        anonId = null
+        anonymousId = null
         coroutineScope.launch {
             repository.clear()
         }
@@ -103,7 +103,7 @@ class ExPlat internal constructor(
     }
 
     private suspend fun fetchAssignments() =
-        repository.fetch(platform, experimentIdentifiers, anonId).fold(
+        repository.fetch(platform, experimentIdentifiers, anonymousId).fold(
             onSuccess = {
                 logger.d("ExPlat: fetching assignments successful with result: $it")
             },
--- a/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
+++ b/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
@@ -24,6 +24,13 @@ class ExPlat internal constructor(
     private val activeVariations = mutableMapOf<String, Variation>()
     private val experimentIdentifiers: List<String> = experiments.map { it.identifier }
 
+    private var anonId: String? = null
+
+    fun configure(anonId: String? = null) {
+        clear()
+        this.anonId = anonId
+    }
+
     /**
      * This returns the current active [Variation] for the provided [Experiment].
      *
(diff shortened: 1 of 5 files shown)


experimentation/src/main/java/com/automattic/android/experimentation/remote/UrlBuilder.kt
```diff
@@ -27,7 +27,9 @@ internal class ExPlatUrlBuilder : UrlBuilder {
             .addPathSegment(platform)
             .apply {
                 experimentNames.forEach { addQueryParameter("experiment_names", it) }
-                addQueryParameter("anon_id", anonymousId.orEmpty())
+                anonymousId?.let {
```
ParaskP7: Minor (🔍): Consider making this a one-liner.

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
```diff
@@ -24,6 +24,13 @@ class ExPlat internal constructor(
     private val activeVariations = mutableMapOf<String, Variation>()
     private val experimentIdentifiers: List<String> = experiments.map { it.identifier }
 
+    private var anonId: String? = null
```
ParaskP7: Suggestion (💡): What if we use the `anonymousId` naming convention instead of this `anonId`, which can mean various other things and not easily understood from the first blink? You did the same elsewhere, I suggest doing it everywhere and only have `anon_id` on the API level the the query parameter.

PS: Same applies to the newly created `test` you added and all references of `anon` there. 😊

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
```diff
@@ -24,6 +24,13 @@ class ExPlat internal constructor(
...
+
+    fun configure(anonId: String? = null) {
```
ParaskP7: Suggestion (💡): Consider adding test(s) for this new public function.

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
```diff
@@ -69,6 +76,7 @@ class ExPlat internal constructor(
     fun clear() {
```
ParaskP7, line 76: Suggestion (💡): Consider adding more test(s) to this new public function. The existing test is only covering that the cache is cleared, but nothing about the `anonymousId` or even that the `activeVariations` is being cleared too.

experimentation/src/test/java/com/automattic/android/experimentation/remote/MockWebServerUrlBuilder.kt
```diff
@@ -0,0 +1,25 @@
+package com.automattic.android.experimentation.remote
+
+import okhttp3.HttpUrl
+import okhttp3.mockwebserver.MockWebServer
+
+internal class MockWebServerUrlBuilder(
```
ParaskP7, line 6: Praise (❤️): Much better! 🥇

ParaskP7 (commented):
Same as with #239 @wzieba , good job and I'll continue reviewing the next PRs until I get to the finish line, then we can start merging everything! 🤞

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
wzieba: Sure 👍 

> PS: Same applies to the newly created test you added and all references of anon there. 😊

But in tests (`ExperimentRestClientTest`) I'd keep `anon` as they're about a query parameter named `anon_id`, not domain "anonymous id", so using `anon` there seems more appropriate.

ParaskP7: Ah, maybe you're right about the tests, it is a `RestClient` test after all! 👍

wzieba:
Hi @ParaskP7 👋 I'm thinking of not adding tests for ExPlat in scope of this PR, as only in one of the next PRs I rewrite most of them. I'll keep in mind your suggestions though.

So, except of tests, does this PR looks good to you for merging?

ParaskP7:
👋 @wzieba !

> Hi @ParaskP7 👋 I'm thinking of not adding tests for ExPlat in scope of this PR, as only in one of the next PRs I rewrite most of them. I'll keep in mind your suggestions though.

Thanks and 👍 !

> So, except of tests, does this PR looks good to you for merging?

Definitely, ready for merging! 🚀

PR #241, opened 2024-08-27
Experimentation without FluxC: example app

wzieba:
Preferably, review:
- #239 
- #240 

first.

### Description

This PR adds possibility to test the `Experimentation` module via the `example` app.

### How to test?

- Go to ExPlat, select a running experiment (preferably with a big number of participants). 
- Enter necessary data: platform to query, experiments name. 
- Generate anon id, tap "setup"
- Test fetching, getting local cache, clearing and see that it works as you'd expect.

Screenshot 2024-08-28 at 09 28 03

### DIFF
--- a/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
+++ b/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
@@ -5,12 +5,16 @@ import com.automattic.android.experimentation.ExPlat.RefreshStrategy.IF_STALE
 import com.automattic.android.experimentation.ExPlat.RefreshStrategy.NEVER
 import com.automattic.android.experimentation.domain.Assignments
 import com.automattic.android.experimentation.domain.AssignmentsValidator
+import com.automattic.android.experimentation.domain.SystemClock
 import com.automattic.android.experimentation.domain.Variation
 import com.automattic.android.experimentation.domain.Variation.Control
+import com.automattic.android.experimentation.local.FileBasedCache
+import com.automattic.android.experimentation.remote.ExperimentRestClient
 import com.automattic.android.experimentation.repository.AssignmentsRepository
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
+import java.io.File
 
 class ExPlat internal constructor(
     private val platform: String,
@@ -113,4 +117,29 @@ class ExPlat internal constructor(
         )
 
     private enum class RefreshStrategy { ALWAYS, IF_STALE, NEVER }
+
+    companion object {
+        fun create(
+            platform: String,
+            experiments: Set<Experiment>,
+            logger: ExperimentLogger,
+            coroutineScope: CoroutineScope,
+            isDebug: Boolean,
+            cacheDir: File,
+        ): ExPlat {
+            val restClient = ExperimentRestClient()
+            val cache = FileBasedCache(cacheDir)
+            val assignmentsRepository = AssignmentsRepository(restClient, cache)
+            val assignmentsValidator = AssignmentsValidator(SystemClock())
+            return ExPlat(
+                platform = platform,
+                experiments = experiments,
+                logger = logger,
+                coroutineScope = coroutineScope,
+                isDebug = isDebug,
+                assignmentsValidator = assignmentsValidator,
+                repository = assignmentsRepository,
+            )
+        }
+    }
 }
--- a/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
+++ b/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
@@ -24,6 +24,13 @@ class ExPlat internal constructor(
     private val activeVariations = mutableMapOf<String, Variation>()
     private val experimentIdentifiers: List<String> = experiments.map { it.identifier }
 
+    private var anonId: String? = null
+
+    fun configure(anonId: String? = null) {
+        clear()
+        this.anonId = anonId
+    }
+
     /**
      * This returns the current active [Variation] for the provided [Experiment].
      *
@@ -69,6 +76,7 @@ class ExPlat internal constructor(
     fun clear() {
         logger.d("ExPlat: clearing cached assignments and active variations")
         activeVariations.clear()
+        anonId = null
         coroutineScope.launch {
             repository.clear()
         }
@@ -95,7 +103,7 @@ class ExPlat internal constructor(
     }
 
     private suspend fun fetchAssignments() =
-        repository.fetch(platform, experimentIdentifiers).fold(
+        repository.fetch(platform, experimentIdentifiers, anonId).fold(
             onSuccess = {
                 logger.d("ExPlat: fetching assignments successful with result: $it")
             },
--- a/sampletracksapp/build.gradle
+++ b/sampletracksapp/build.gradle
@@ -47,6 +47,7 @@ dependencies {
 
     // For performance showcase
     implementation "com.squareup.okhttp3:okhttp:$squareupOkhttpVersion"
+    implementation project(':experimentation')
 }
 
 dependencyAnalysis {
--- a/sampletracksapp/build.gradle
+++ b/sampletracksapp/build.gradle
@@ -41,13 +41,13 @@ android {
 dependencies {
     implementation project(":AutomatticTracks")
     implementation project(":crashlogging")
+    implementation project(':experimentation')
     implementation "androidx.appcompat:appcompat:$androidxAppcompatVersion"
     implementation "com.google.android.material:material:$googleMaterialVersion"
     implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:$kotlinxCoroutinesVersion"
 
     // For performance showcase
     implementation "com.squareup.okhttp3:okhttp:$squareupOkhttpVersion"
-    implementation project(':experimentation')
 }
 
 dependencyAnalysis {
--- a/sampletracksapp/build.gradle
+++ b/sampletracksapp/build.gradle
@@ -45,6 +45,7 @@ dependencies {
     implementation "androidx.appcompat:appcompat:$androidxAppcompatVersion"
     implementation "com.google.android.material:material:$googleMaterialVersion"
     implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:$kotlinxCoroutinesVersion"
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:$kotlinxCoroutinesVersion"
 
     // For performance showcase
     implementation "com.squareup.okhttp3:okhttp:$squareupOkhttpVersion"
--- a/sampletracksapp/src/main/java/com/example/sampletracksapp/ExperimentationDialogFragment.kt
+++ b/sampletracksapp/src/main/java/com/example/sampletracksapp/ExperimentationDialogFragment.kt
@@ -13,6 +13,7 @@ import com.example.sampletracksapp.databinding.DialogExperimentationBinding
 import kotlinx.coroutines.GlobalScope
 import java.io.File
 import java.util.UUID
+import kotlinx.coroutines.DelicateCoroutinesApi
 
 class ExperimentationDialogFragment : DialogFragment() {
 
@@ -43,7 +44,7 @@ class ExperimentationDialogFragment : DialogFragment() {
                         }
                     }?.toSet().orEmpty(),
                     cacheDir = cacheDir,
-                    coroutineScope = GlobalScope,
+                    coroutineScope = @Suppress("OPT_IN_USAGE") GlobalScope,
                     isDebug = true,
                     logger = object : ExperimentLogger {
                         override fun d(message: String) {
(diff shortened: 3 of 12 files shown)


experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
```diff
@@ -113,4 +117,29 @@ class ExPlat internal constructor(
...
+            logger: ExperimentLogger,
+            coroutineScope: CoroutineScope,
+            isDebug: Boolean,
+            cacheDir: File,
+        ): ExPlat {
+            val restClient = ExperimentRestClient()
+            val cache = FileBasedCache(cacheDir)
+            val assignmentsRepository = AssignmentsRepository(restClient, cache)
+            val assignmentsValidator = AssignmentsValidator(SystemClock())
+            return ExPlat(
```
ParaskP7, line 134: Suggestion (💡): How about removing the need for local parameters, them not being shared, thus not that much needed?

```
            return ExPlat(
                platform = platform,
                experiments = experiments,
                logger = logger,
                coroutineScope = coroutineScope,
                isDebug = isDebug,
                assignmentsValidator = AssignmentsValidator(SystemClock()),
                repository = AssignmentsRepository(
                    ExperimentRestClient(),
                    FileBasedCache(cacheDir)
                ),
            )
```

sampletracksapp/build.gradle
```diff
@@ -47,6 +47,7 @@ dependencies {
 
     // For performance showcase
     implementation "com.squareup.okhttp3:okhttp:$squareupOkhttpVersion"
+    implementation project(':experimentation')
```
ParaskP7: Minor (🔍): Consider grouping all `project` related dependencies together on top.

PS: I usually prefer the below grouping:
1. `project` dependencies
2. `lib` dependencies
3. `debug` dependencies
4. `main` dependencies
5. `test` dependencies
6. `ui test`  dependencies
7. `other` dependencies

Just so I know what is happening where...

sampletracksapp/src/main/java/com/example/sampletracksapp/ExperimentationDialogFragment.kt
```diff
@@ -0,0 +1,91 @@
...
+import androidx.fragment.app.DialogFragment
+import com.automattic.android.experimentation.ExPlat
+import com.automattic.android.experimentation.Experiment
+import com.automattic.android.experimentation.ExperimentLogger
+import com.example.sampletracksapp.databinding.DialogExperimentationBinding
+import kotlinx.coroutines.GlobalScope
+import java.io.File
+import java.util.UUID
+
+class ExperimentationDialogFragment : DialogFragment() {
```
ParaskP7, line 27: 1. Praise (❤️): Thanks for extracting this set of experimentation testing views in its own screen.
2. Suggestion (💡): Consider doing the same for the tracks testings view and extracting them into its own screen, if that's not asking too much... 😊

sampletracksapp/src/main/java/com/example/sampletracksapp/ExperimentationDialogFragment.kt
```diff
@@ -0,0 +1,91 @@
...
+        DialogExperimentationBinding.inflate(inflater, container, false).apply {
+            setup.setOnClickListener {
+                exPlat = ExPlat.create(
+                    platform = platform.text.toString(),
+                    experiments = experiments.text?.toString()?.split(",")?.map {
+                        object : Experiment {
+                            override val identifier: String = it
+                        }
+                    }?.toSet().orEmpty(),
+                    cacheDir = context!!.cacheDir,
```
ParaskP7: Minor (🔍): Consider using `requireContext()` to avoid the `!!` and us needing to worry about the warning in this file.

PS: Same applies further below with another such change (see `getCache.setOnClickListener { ... }`).

sampletracksapp/src/main/java/com/example/sampletracksapp/ExperimentationDialogFragment.kt
```diff
@@ -0,0 +1,91 @@
...
+                    coroutineScope = GlobalScope,
```
ParaskP7: Minor (🔍): Consider suppressing this with a `@Suppress("OPT_IN_USAGE")` on the statement itself.

ParaskP7 (commented):
Same as with #239 and #240 @wzieba , great work on creating this testing screen, very helpful, and I'll continue reviewing the next PRs until I get to the finish line, then we can start merging everything! 🤞

sampletracksapp/src/main/java/com/example/sampletracksapp/ExperimentationDialogFragment.kt
```diff
@@ -0,0 +1,91 @@
...
+    override fun onStart() {
+        super.onStart()
+        dialog?.let {
+            val width = ViewGroup.LayoutParams.MATCH_PARENT
+            val height = ViewGroup.LayoutParams.MATCH_PARENT
+            it.window?.setLayout(width, height)
+        }
+    }
+
+    override fun onCreateView(
```
ParaskP7, line 40: FYI: This is the second time I write this comment (😭) as I forgot to save it, then submitted the review without saving it, which caused me to lose it... 😞 

-----

Question (❓): While testing this I was only getting assignments with an empty list of `variations`. I remember you showcasing that to us the other day and that it works, it is just that I couldn't make it show assignments with `variations`. I used the following:

- Platform to query: `wpandroid`
- Experiment names: Copied the `name` from [ExPlat](https://experiments.a8c.com/experiments)
- Anon ID: Tried with and without that
- How: First clicked `SETUP` then `FETCH`

No sure if I am just unlucky, doing something wrong, or it not working as I expected... 🤷 

-----

`EXTRAS`:

1. Suggestion (💡): To show the output of the result of our actions in a `TextView` below the buttons, instead of jumping to `LogCat`, just like it is done on `FluxC`, and help testers test.
2. Suggestion (💡): To make `Platform to Query` mandatory before hitting `SETUP`, thus disabling this button at first. Same applies to `Experiment names`.
3. Suggestion (💡): To mark `Anon ID` optional so that testers know this is not required.
4. Suggestion (💡): To disable the `FETCH` button until `SETUP` is triggered.
5. Suggestion (💡): To move the `GET CACHE` and `CLEAR CACHE` button on a separate row, below the `FETCH` and `SETUP` button, them being a "somehow" secondary testing buttons.

PS: Please ignore any of those above, I just wanted to put those idea here for you, just in case you would have liked applying some. We can also do all that later-on.

wzieba: > Platform to query

You have to choose the correct platform - I see on ExPlat there's no experiments for `wpandroid` right now. It's a third column on the table :)

ParaskP7: Ah @wzieba , you're right, I was looking for this but within a specific experiment page (ie. 21918-explat-experiment) and couldn't find it... It was in front of my eyes and you also showed that to us! 😭

FYI: Btw, I just tested that again with a few such experiment names, using the correct platform name this time and I never got an `assignment` with a non-empty list of `variations`... 🤷

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
wzieba: Sure 👍. Let's do this in the scope of #243 to reduce number of conflicts: https://github.com/Automattic/Automattic-Tracks-Android/pull/243/files#r1738128236

ParaskP7: I agree! 👍

sampletracksapp/src/main/java/com/example/sampletracksapp/ExperimentationDialogFragment.kt
wzieba: > FYI: Btw, I just tested that again with a few such experiment names, using the correct platform name this time and I never got an assignment with a non-empty list of variations... 🤷

This can be the case of an experiment not working with `anonId`, but maybe with a bearer token identification. I've asked a clarifying question internally, p1725007206945009-slack-C7HH3V5AS

But for now, please check the currently running `calypso` experiment with higher number of participants - this one should return some variations (I just tested it)

wzieba: > EXTRAS:

Initially, I wasn't sure if it's worth investing time into this, but maybe you're right - better example app maybe can help with understanding the SDK in the future.

I've applied all the suggestions, take a look!

sampletracksapp/src/main/java/com/example/sampletracksapp/ExperimentationDialogFragment.kt
wzieba: > Suggestion (💡): Consider doing the same for the tracks testings view and extracting them into its own screen, if that's not asking too much... 😊

Thanks, but I'd prefer to keep it in a separate PR (I believe you meant "crashlogging/Sentry" testing view, as currently example app doesn't have testing views for Tracks)

sampletracksapp/src/main/java/com/example/sampletracksapp/ExperimentationDialogFragment.kt
ParaskP7: 👋 @wzieba and thanks for the reply!

> This can be the case of an experiment not working with anonId, but maybe with a bearer token identification. I've asked a clarifying question internally, p1725007206945009-slack-C7HH3V5AS
> 
> But for now, please check the currently running calypso experiment with higher number of participants - this one should return some variations (I just tested it)

Thanks for the message on Slack! 🙇 

Btw, I did try with and without the `anonId` yesterday, and did try this `calypso` experiment you suggested, did work, was only getting assignments without any variations... 🤷 

But, I just tried again, and although the first time, without `anonId` I wasn't getting any variations, when I tried a second time and SETUP the experiment with an `anonId` it start working, the first time I see an assignment with a variation: 🎉 

`variations={calypso_disable_site_assembler=Treatment(name=treatment_disable_onboarding)}`

So, all good I think, I was just unlucky (I guess) yesterday all day, not being able to get any variations... 😅 

-----

> Initially, I wasn't sure if it's worth investing time into this, but maybe you're right - better example app maybe can help with understanding the SDK in the future.
> 
> I've applied all the suggestions, take a look!

Yea, I only suggested those because I thought them to be quick wins, thanks for applying the suggestion, taking a look se speak! 🥇

ParaskP7: This awesome @wzieba , I just tested the improvements and that looks so much better! 🥇 

Few extra quick suggestions to take this to 100%:
1. Making disabling the `GENERATE` too when `SETUP` hasn't happened yet. This is because of the `exPlat?.clear()` which will not work when hitting generate, but maybe someone will expect that. PS: Same apply for `CLEAR CACHE` because maybe it will become misleading/frustrating to testers thinking this works. Only `GET CACHE` does not depend on `exPlat` being initialized.
2. Adding the result of `GET CACHE` to the output too. This way it will become evident whether `CLEAR CACHE` worked or not.
3. When `GET CACHE` doesn't exist, let's output that too, both on screen and in the `LogCat` logs.

sampletracksapp/src/main/java/com/example/sampletracksapp/ExperimentationDialogFragment.kt
ParaskP7: 😅 Yea, yea, I meant for `Crash Logging/Sentry` testing view, you're right! 💯

sampletracksapp/src/main/java/com/example/sampletracksapp/ExperimentationDialogFragment.kt
wzieba: Thanks, good suggestions 👍 I applied all the fixes.

And I found an issue, that addressing might help in testing: `generateAnonId` wasn't providing the anonymous id to `ExPlat` 🫣 Fixed in 17fd0e0

This proves though, that `anonId` is not a requirement for some experiments.

ParaskP7: > Thanks, good suggestions 👍 I applied all the fixes.

This is awesome, I never had to open the `LogCat` during this test iteration!

> And I found an issue, that addressing might help in testing: generateAnonId wasn't providing the anonymous id to ExPlat 🫣 Fixed in https://github.com/Automattic/Automattic-Tracks-Android/commit/17fd0e05b2697048015f9a8915b9c7289fa51afe

True, when reviewing I was thinking that you expect testers to hit `SETUP` afterwards, but with this change this is not needed! 💯 

> This proves though, that anonId is not a requirement for some experiments.

I guess so! 😅 

-----

This too I think it is ready to merge! 🚀

wzieba: > True, when reviewing I was thinking that you expect testers to hit SETUP afterwards, but with this change this is not needed! 💯

I think this was initially my idea, but I think it was not very clear (it wasn't for me after 1 day 😄 )

ParaskP7: > (it wasn't for me after 1 day 😄 )

Then it is 100% the right call! 🤣 🤣 🤣

PR #242, opened 2024-08-28
Improve threading and test suite

wzieba:
### Description

This PR aims to:
- add some improvements to threading by removing `runBlocking` call in `getAssignments` method
- improve the test suite (https://github.com/Automattic/Automattic-Tracks-Android/pull/239#discussion_r1732906646)
- fix a quick in public API design of requesting cache refresh when getting assignments

### To test
This PR introduces a robust test suite. Still, you can perform some smoke tests if you'd like following instructions from #241

### DIFF
--- a/experimentation/build.gradle
+++ b/experimentation/build.gradle
@@ -49,9 +49,6 @@ dependencies {
     testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:$kotlinxCoroutinesVersion"
     testImplementation "junit:junit:$junitVersion"
     testImplementation "org.assertj:assertj-core:$assertjVersion"
-    testImplementation "org.mockito.kotlin:mockito-kotlin:$mockitoKotlinVersion"
-    testImplementation "org.mockito:mockito-inline:$mockitoInlineVersion"
-
     testImplementation "com.squareup.okhttp3:mockwebserver:$okHttp"
 }
 
--- a/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
+++ b/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
@@ -48,7 +48,7 @@ class ExPlat internal constructor(
      * If the provided [Experiment] was not included in [experiments], then [Control] is returned.
      * If [isDebug] is `true`, an [IllegalArgumentException] is thrown instead.
      */
-    fun getVariation(experiment: Experiment): Variation? {
+    fun getVariation(experiment: Experiment): Variation {
         val experimentIdentifier = experiment.identifier
         if (!experimentIdentifiers.contains(experimentIdentifier)) {
             val message = "ExPlat: experiment not found: \"${experimentIdentifier}\"! " +
--- a/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
+++ b/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
@@ -11,9 +11,10 @@ import com.automattic.android.experimentation.domain.Variation.Control
 import com.automattic.android.experimentation.local.FileBasedCache
 import com.automattic.android.experimentation.remote.ExperimentRestClient
 import com.automattic.android.experimentation.repository.AssignmentsRepository
+import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.runBlocking
 import java.io.File
 
 class ExPlat internal constructor(
@@ -93,8 +94,7 @@ class ExPlat internal constructor(
     }
 
     private fun getAssignments(refreshStrategy: RefreshStrategy): Assignments {
-        val cachedAssignments: Assignments =
-            runBlocking { repository.getCached() } ?: Assignments(emptyMap(), 0, 0)
+        val cachedAssignments: Assignments = repository.getCached() ?: Assignments(emptyMap(), 0, -1)
         if (refreshStrategy == ALWAYS || (
                 refreshStrategy == IF_STALE && assignmentsValidator.isStale(
                     cachedAssignments,
@@ -124,11 +124,12 @@ class ExPlat internal constructor(
             experiments: Set<Experiment>,
             logger: ExperimentLogger,
             coroutineScope: CoroutineScope,
+            dispatcher: CoroutineDispatcher = Dispatchers.IO,
             isDebug: Boolean,
             cacheDir: File,
         ): ExPlat {
-            val restClient = ExperimentRestClient()
-            val cache = FileBasedCache(cacheDir)
+            val restClient = ExperimentRestClient(dispatcher = dispatcher)
+            val cache = FileBasedCache(cacheDir, dispatcher = dispatcher, scope = coroutineScope)
             val assignmentsRepository = AssignmentsRepository(restClient, cache)
             val assignmentsValidator = AssignmentsValidator(SystemClock())
             return ExPlat(
--- a/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
+++ b/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
@@ -48,10 +48,7 @@ class ExPlat internal constructor(
      * If the provided [Experiment] was not included in [experiments], then [Control] is returned.
      * If [isDebug] is `true`, an [IllegalArgumentException] is thrown instead.
      */
-    fun getVariation(
-        experiment: Experiment,
-        shouldRefreshIfStale: Boolean = false,
-    ): Variation {
+    fun getVariation(experiment: Experiment): Variation {
         val experimentIdentifier = experiment.identifier
         if (!experimentIdentifiers.contains(experimentIdentifier)) {
             val message = "ExPlat: experiment not found: \"${experimentIdentifier}\"! " +
@@ -65,7 +62,7 @@ class ExPlat internal constructor(
             }
         }
         return activeVariations.getOrPut(experimentIdentifier) {
-            getAssignments(if (shouldRefreshIfStale) IF_STALE else NEVER)
+            getAssignments(NEVER)
                 .variations[experimentIdentifier] ?: Control
         }
     }
--- a/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
+++ b/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
@@ -48,7 +48,7 @@ class ExPlat internal constructor(
      * If the provided [Experiment] was not included in [experiments], then [Control] is returned.
      * If [isDebug] is `true`, an [IllegalArgumentException] is thrown instead.
      */
-    fun getVariation(experiment: Experiment): Variation {
+    fun getVariation(experiment: Experiment): Variation? {
         val experimentIdentifier = experiment.identifier
         if (!experimentIdentifiers.contains(experimentIdentifier)) {
             val message = "ExPlat: experiment not found: \"${experimentIdentifier}\"! " +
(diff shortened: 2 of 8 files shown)


experimentation/build.gradle
```diff
@@ -49,9 +49,6 @@ dependencies {
     testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:$kotlinxCoroutinesVersion"
     testImplementation "junit:junit:$junitVersion"
     testImplementation "org.assertj:assertj-core:$assertjVersion"
-    testImplementation "org.mockito.kotlin:mockito-kotlin:$mockitoKotlinVersion"
-    testImplementation "org.mockito:mockito-inline:$mockitoInlineVersion"
```
ParaskP7, line 52: Praise (❤️): Woohoo, gone, bye-bye, so-long, hasta-la-svista! 👋

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
```diff
@@ -47,10 +48,7 @@ class ExPlat internal constructor(
      * If the provided [Experiment] was not included in [experiments], then [Control] is returned.
      * If [isDebug] is `true`, an [IllegalArgumentException] is thrown instead.
      */
-    fun getVariation(
-        experiment: Experiment,
-        shouldRefreshIfStale: Boolean = false,
```
ParaskP7, line 52: Think (🤔): Interestanto, I'll wait for how the APIs ends-up being in the end, also, we might want to update the KDoc for this function.

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
```diff
@@ -64,7 +62,7 @@ class ExPlat internal constructor(
             }
         }
         return activeVariations.getOrPut(experimentIdentifier) {
-            getAssignments(if (shouldRefreshIfStale) IF_STALE else NEVER)
+            getAssignments(NEVER)
```
ParaskP7: Minor (🔍): This can now become a one-liner.

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
```diff
@@ -93,8 +91,7 @@ class ExPlat internal constructor(
     }
 
     private fun getAssignments(refreshStrategy: RefreshStrategy): Assignments {
-        val cachedAssignments: Assignments =
-            runBlocking { repository.getCached() } ?: Assignments(emptyMap(), 0, 0)
```
ParaskP7, line 95: Praise (❤️): Nicely done with the in-memory value, I like! 🥇

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
```diff
@@ -93,8 +91,7 @@ class ExPlat internal constructor(
...
+        val cachedAssignments: Assignments = repository.getCached() ?: Assignments(emptyMap(), 0, -1)
```
ParaskP7: Question (❓): What's with the `-1` change? 🤔

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
```diff
@@ -124,11 +121,12 @@ class ExPlat internal constructor(
             experiments: Set<Experiment>,
             logger: ExperimentLogger,
             coroutineScope: CoroutineScope,
+            dispatcher: CoroutineDispatcher = Dispatchers.IO,
```
ParaskP7, line 119: Suggestion (🔍): I suggest moving `dispatcher` and `coroutineScope` at the end of the function's signatures, them becoming the last parameters in it.

experimentation/src/test/java/com/automattic/android/experimentation/local/FileBasedCacheTest.kt
```diff
@@ -52,13 +54,13 @@ internal class FileBasedCacheTest {
...
-        val sut = fileBasedCache()
+        val sut = fileBasedCache(this)
 
         sut.clear()
     }
 
-    private fun fileBasedCache() =
-        FileBasedCache(cacheDir = createTempDirectory().toFile())
+    private fun fileBasedCache(scope: TestScope) =
+        FileBasedCache(cacheDir = createTempDirectory().toFile(), dispatcher = StandardTestDispatcher(scope.testScheduler), scope = scope)
```
ParaskP7: Minor (🔍): Consider making this a bit more readable:

```
    private fun fileBasedCache(scope: TestScope) = FileBasedCache(
        cacheDir = createTempDirectory().toFile(),
        dispatcher = StandardTestDispatcher(scope.testScheduler),
        scope = scope
    )
```

experimentation/src/test/java/com/automattic/android/experimentation/remote/ExperimentRestClientTest.kt
```diff
@@ -13,24 +15,15 @@ import okhttp3.mockwebserver.RecordedRequest
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertTrue
-import org.junit.Before
 import org.junit.Test
 
 internal class ExperimentRestClientTest {
 
     private val server: MockWebServer = MockWebServer()
```
ParaskP7, line 22: Praise (❤️): Love the new non-mocked tests! 🥇

experimentation/src/test/java/com/automattic/android/experimentation/ExPlatTest.kt
```diff
@@ -2,267 +2,196 @@ package com.automattic.android.experimentation
...
+        )
     }
 
     @Test
-    fun `refreshIfNeeded does not fetch assignments if cache is fresh`() = runBlockingTest {
-        setupAssignments(cachedAssignments = buildAssignments(isStale = false), fetchedAssignments = buildAssignments())
+    fun `refreshing in case of fresh cache doesn't fetch new assignments`() = runTest {
+        enqueueSuccessfulNetworkResponse()
+        val exPlat = createExPlat(clock = { 3599 })
+        tempCache.saveAssignments(testAssignment.copy(timeToLive = 3600))
```
ParaskP7, line 61: Praise (❤️): Nicely done!

experimentation/src/test/java/com/automattic/android/experimentation/ExPlatTest.kt
```diff
@@ -2,267 +2,196 @@ package com.automattic.android.experimentation
...
+        assertThat(tempCache.latest).isNull()
+    }
+
+    /*
+    This scenario is about returning the same variation during a single session.
+    We don't want SDK consumers to mix variations during the same session,
+    as it could lead to unexpected behavior.
+     */
+    @Test
+    fun `getting variation for the second time returns the same value, even if cache was updated`() =
```
ParaskP7, line 102: 1. Praise (❤️): Thanks for this test, it helps further clarifying few API decisions.
2. Minor (🔍): Btw, you could move `runTest {` on the same line to the `fun` itself and make this test more readable, give it more horizontal space.

experimentation/src/test/java/com/automattic/android/experimentation/ExPlatTest.kt
```diff
@@ -2,267 +2,196 @@ package com.automattic.android.experimentation
...
     @Test
-    fun `getVariation does not fetch assignments if cache is stale but shouldRefreshIfStale is false`() = runBlockingTest {
-        setupAssignments(cachedAssignments = null, fetchedAssignments = buildAssignments())
-
-        exPlat.getVariation(dummyExperiment, shouldRefreshIfStale = false)
+    fun `in debug, getting variation throws an exception experiment was not found`() = runTest {
+        val exPlat = createExPlat(experiments = emptySet())
 
-        verify(restClient, never()).fetchAssignments(eq(platform), any(), anyOrNull())
+        assertThatThrownBy {
```
ParaskP7, line 128: Praise (❤️): TIL

experimentation/src/test/java/com/automattic/android/experimentation/ExPlatTest.kt
```diff
@@ -2,267 +2,196 @@ package com.automattic.android.experimentation
...
-
-        val firstVariation = exPlat.getVariation(dummyExperiment, shouldRefreshIfStale = false)
-        assertThat(firstVariation).isEqualTo(controlVariation)
-
-        exPlat.forceRefresh()
-
-        setupAssignments(cachedAssignments = treatmentAssignments, fetchedAssignments = treatmentAssignments)
-
-        val secondVariation = exPlat.getVariation(dummyExperiment, shouldRefreshIfStale = false)
-        assertThat(secondVariation).isEqualTo(controlVariation)
```
ParaskP7, line 181: Suggestion (💡): I usually split such code with an extra `/* HELPER */` grouping comment, which means, see helper functions below.

ParaskP7 (commented):
Same as with #239, #240 and #241 @wzieba , awesome work on the testing side of things, very helpful, and I'll continue reviewing the final PR, I am very close to the finish line, then we can start merging everything! 🤞

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
ParaskP7: UPDATE: I am seeing that the KDoc thingy is happening on https://github.com/Automattic/Automattic-Tracks-Android/pull/243/commits/990b3a1d78f2f138cafa05c839694c32695b639e , marking this as resolved.

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
wzieba: Added to: https://github.com/Automattic/Automattic-Tracks-Android/pull/243#discussion_r1738128236

experimentation/src/test/java/com/automattic/android/experimentation/ExPlatTest.kt
wzieba: > Btw, you could move runTest { on the same line

I wish, but then auto-formatting moves it back. I didn't want to spend much time on fixing the rule, though

ParaskP7: Ah, I have a default line length set at 120 before hard wrapping, maybe that's why, yea, please ignore then... 🤷

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
wzieba: TLDR; I replaced this with `getAssignments` possible to return `null` now. I think it's much more readable now, and makes more sense: if there's no cache, it's better to return `null` than an `Assignment` with made up values. 98bba0a58977c64dac88429d5e1f07c3ad0d380a

Context here was that I wanted this Assignment to be `stale` when `Clock` is `0` (in tests, making things more readable). It was also ok for my thinking process to make `fetchedAt` value a minus value (that doesn't make sense) to signal that it's not a valid object. But as I moved to use `null`, this doesn't matter.

wzieba:
@ParaskP7 I think this PR is ready to re-review. Re tests we discussed in https://github.com/Automattic/Automattic-Tracks-Android/pull/240#discussion_r1736152248 - I think I'll add it rather in #243 as there are some changes to tests there, so I don't want to work unnecessarily addressing conflicts

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
ParaskP7: Great, thanks for this change @wzieba , I agree, this seems much more robust to me! 💯

ParaskP7:
Reviewed and everything looks good to me, feel free to merge! ✅ 

PS: Just one note from me, maybe we should consider testing the publicly available `getVariation(...)` call, via the example app, as I was surprised that the change to making it nullable didn't have any effect their too... 🤔

wzieba:
Sorry, my bad. Actually `getVariation` is not changing signature - in case of empty assignment, we return `Control`

ParaskP7:
> Sorry, my bad. Actually getVariation is not changing signature - in case of empty assignment, we return Control

Ah, right, true! 💯

PR #243, opened 2024-08-29
Experiments public contract

wzieba:
### Description

This PR introduces improvements to the public contract:
- extracts an interface: `VariationsRepository`, so the clients can easily double this in tests + KDoc comments
- introduces a behavior change: refreshing cache on SDK initialization, as I believe this is something that every client would do on initialization anyway
- simplifies `refresh` API (no need for two methods)
- adds `README.md` with an explanation about how to use the SDK

### Testing

The only behavior change has been covered  with automated tests. It's still possible to smoke test using instructions from #241 , if necessary.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -30,10 +30,18 @@ steps:
       cp gradle.properties-example gradle.properties
       ./gradlew test
 
+  - label: "API verification"
+    key: "api"
+    plugins: [$CI_TOOLKIT]
+    command: |
+      cp gradle.properties-example gradle.properties
+      ./gradlew :experimentation:apiCheck
+
   - label: "Publish {{matrix}} to S3"
     depends_on:
       - "lint"
       - "test"
+      - "api""
     plugins: [$CI_TOOLKIT]
     matrix:
       - AutomatticTracks
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -41,7 +41,7 @@ steps:
     depends_on:
       - "lint"
       - "test"
-      - "api""
+      - "api"
     plugins: [$CI_TOOLKIT]
     matrix:
       - AutomatticTracks
--- a/experimentation/README.md
+++ b/experimentation/README.md
@@ -1,0 +1,63 @@
+# Experimentation
+
+## Introduction
+
+This module provides an SDK to communicate with Automattic-internal A/B testing service: ExPlat.
+
+## Coordinates
+
+```groovy
+dependencies {
+    implementation 'com.automattic.tracks:experimentation:{recent_release}'
+}
+```
+
+## Usage
+
+### Initialization
+
+```kotlin
+val experiments = setOf(
+    Experiment("experiment_1"),
+    Experiment("experiment_2")
+)
+
+// If injected in a DI container, the repository should be a singleton
+val repository: VariationsRepository = VariationsRepository.create(
+    experiments = experiments
+    /* see KDoc documentation for detailed description of parameters */
+)
+repository.configure(anonId = "currently_logged_in_user_id_or_random_uuid")
+```
+
+### Getting variation
+
+```kotlin
+// New variations will be available on the next application session. See KDoc for `VariationsRepository#getVariation`
+val variation = repository.getVariation("experiment_1")
+
+when (variation) {
+    is Control -> {
+        // continue with control group
+    }
+    is Treatment -> {
+        println("Treatment group: ${variation.value}")
+        // apply treatment changes
+    }
+}
+```
+
+### User logs out
+
+```kotlin
+repository.clean()
+```
+
+### Different user logs in
+
+```kotlin
+repository.configure(anonId = "new_user_id")
+```
+
+## Testing
+The SDK offers `VariantionsRepository` interface, which can be doubled in tests.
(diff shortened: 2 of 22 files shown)


experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
```diff
@@ -5,50 +5,38 @@ import com.automattic.android.experimentation.ExPlat.RefreshStrategy.IF_STALE
...
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
-import java.io.File
 
-class ExPlat internal constructor(
+public class ExPlat internal constructor(
     private val platform: String,
-    private val experiments: Set<Experiment>,
+    experiments: Set<Experiment>,
```
wzieba, line 13: I wonder: is there a value in providing experiments in the constructor and then checking for its existence in `getVariation`? I don't see good benefits 🤔

.buildkite/pipeline.yml
```diff
@@ -30,10 +30,18 @@ steps:
       cp gradle.properties-example gradle.properties
       ./gradlew test
 
+  - label: "API verification"
+    key: "api"
+    plugins: [$CI_TOOLKIT]
+    command: |
+      cp gradle.properties-example gradle.properties
+      ./gradlew :experimentation:apiCheck
```
ParaskP7, line 38: Praise (❤️): Awesome, love it, thanks for adding the `binary-compatibility-validator` plugin into this repo, let's start using this more and more within this project and other repos as well! 🥇

experimentation/README.md
```diff
@@ -0,0 +1,63 @@
+# Experimentation
```
ParaskP7, line 1: Praise (❤️): Thanks for creating a `README.md` to help with the API usage!

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
```diff
@@ -5,50 +5,38 @@ import com.automattic.android.experimentation.ExPlat.RefreshStrategy.IF_STALE
...
+) : VariationsRepository {
     private val activeVariations = mutableMapOf<String, Variation>()
     private val experimentIdentifiers: List<String> = experiments.map { it.identifier }
 
     private var anonId: String? = null
 
-    fun configure(anonId: String? = null) {
+    init {
+        coroutineScope.launch {
+            refresh()
```
ParaskP7: Question (❓): I think this is a good move, but how about when users would like to also use `configure(...)` to setup ExPlat with an anonymous id, they will then need to trigger `refresh()` manually once more, right, or, how will they better use the API then? 🤔

experimentation/src/main/java/com/automattic/android/experimentation/VariationsRepository.kt
```diff
@@ -0,0 +1,85 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import java.io.File
+
+/**
+ * Repository for accessing [Variation]s for [Experiment]s.
+ *
+ * Should be a Singleton, as it keeps track of the "active" [Variation]s. See [getVariation] for more details.
+ */
+public interface VariationsRepository {
```
ParaskP7, line 20: Praise (❤️): Love it! 🥇

ParaskP7 (commented):
And with that final review, I think I am done with all the review @wzieba , kudos on this work, it will help our product teams setup and start using [ExPlat](https://experiments.a8c.com/experiments) without much overhead, and mostly, without `FluxC`! 🎉 🥇 🚀 

Let's start resolving the comments one-by-one and merge everything to `trunk` when ready, just let me know how to help you from here! 💯

experimentation/src/main/java/com/automattic/android/experimentation/VariationsRepository.kt
```diff
@@ -0,0 +1,85 @@
...
+            val assignmentsValidator = AssignmentsValidator(SystemClock())
+            return ExPlat(
+                platform = platform,
+                experiments = experiments,
+                logger = logger,
+                coroutineScope = coroutineScope,
+                isDebug = isDebug,
+                assignmentsValidator = assignmentsValidator,
+                repository = assignmentsRepository,
+            )
```
wzieba: - [ ] https://github.com/Automattic/Automattic-Tracks-Android/pull/241#discussion_r1736393960
- [ ] https://github.com/Automattic/Automattic-Tracks-Android/pull/242#discussion_r1736610369

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
wzieba: True, this was not the best idea. I learned that `anonId` is actually a mandatory information ( p1725007206945009-slack-C7HH3V5AS) and I proposed changes to this initialization methods in #245

ParaskP7: Thanks for all this work @wzieba , I followed all the conversation on anonymous id needing to become mandatory (and all), plus did a quick review on #245 , the changes makes sense to me (also, thanks @iangmaia for the review)! 🚀

PR #244, opened 2024-08-30
feature branch: refactor `experimentation` to not use FluxC

wzieba:
Previously merged:
- #237 
- #238 

Contains:
- #239 
- #240 
- #241 
- #242 
- #243 
- #245 
- #246

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -30,10 +30,18 @@ steps:
       cp gradle.properties-example gradle.properties
       ./gradlew test
 
+  - label: "API verification"
+    key: "api"
+    plugins: [$CI_TOOLKIT]
+    command: |
+      cp gradle.properties-example gradle.properties
+      ./gradlew :experimentation:apiCheck
+
   - label: "Publish {{matrix}} to S3"
     depends_on:
       - "lint"
       - "test"
+      - "api""
     plugins: [$CI_TOOLKIT]
     matrix:
       - AutomatticTracks
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -41,7 +41,7 @@ steps:
     depends_on:
       - "lint"
       - "test"
-      - "api""
+      - "api"
     plugins: [$CI_TOOLKIT]
     matrix:
       - AutomatticTracks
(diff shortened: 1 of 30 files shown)


wzieba:
### Review

hi @necronet 👋 . This PR consist several smaller PRs and some of the code is Android platform-specific, so I thought about leaving some context to make the review easier.

- The main API contract class is `VariationsRepository`. See the attached code comments.
- To see how the SDK should be integrated, check out `experimentation/README.md` file and/or this basic integration POC https://github.com/Automattic/pocket-casts-android/pull/2800
- I've build the sample app, so you don't have to install all Android tooling. I'll send it on Slack

necronet (approved):
LGTM I checked through the code and the test app.  Making the right call and with the right parameters.

Screenshot 2024-09-06 at 16 27 13

As requested the `fluxc` and `wellsql` dependencies have been removed, and these changes now use a filebased system for caching the variations in the Repository class provided to ExPlat.  

I'd enforce a non-empty anonymous id throught this sdk though instead of `anonymousId?.let { addQueryParameter("anon_id", it) }` but I am not sure if it will break something else, as this something that we discussed it should be fine, API can handle this scenario quite well.

wzieba:
Thank you for the review @necronet !

> I'd enforce a non-empty anonymous id throught this sdk

That's a good point, but this is how the proposed change works now:

https://github.com/Automattic/Automattic-Tracks-Android/blob/87dfdcf59774635393274d0359393b718207156a/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt#L35-L38

https://github.com/Automattic/Automattic-Tracks-Android/blob/87dfdcf59774635393274d0359393b718207156a/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt#L59-L66

When `anonymousId` is not provided and variation is requested, we return `Control` or crash the app (if it's `failFast`, usually in debug).

The `anonymousId?.let { addQueryParameter("anon_id", it) }` part you mention is solely for safety if something unexpected might've happened with `anonymousId`.

wzieba:
> I'd enforce a **non-empty** anonymous id 

After reading the comment again, I've enhanced this guard with a `isNullOrBlank` check (instead of only `null` check).

PR #245, opened 2024-09-02
Make anon id mandatory. Simplify API, remove `refresh` 

wzieba:
### Description

After learning that "anon id" is a necessary piece when requesting experiments assignments (internal source: p1725007206945009-slack-C7HH3V5AS), I've decided to update the public contract to reflect this change and inform SDK consumers early, that anon id is necessary (`initialization`) rather than optional (`configure`)

Furthermore, this PR removes `refresh` method. To my understanding, this was leaking implementation detail. We shouldn't bother SDK consumers when and how cache should be updated - the SDK should handle its own cache.

### How to review

I recommend checking this PR commit by commit and check commit descriptions - I've tried to add more context for each change there and tell a story of changes.

### How to test

The changes are covered with automated unit tests. You can run an example app and follow instructions from #241 but it's not necessary - especially as this PR targets a feature branch.

### DIFF
--- a/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
+++ b/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
@@ -27,9 +27,7 @@ public class ExPlat internal constructor(
 
     override fun initialize(anonymousId: String) {
         this.anonymousId = anonymousId
-        coroutineScope.launch {
-            refresh()
-        }
+        refresh(IF_STALE)
     }
 
     override fun getVariation(experiment: Experiment): Variation {
@@ -48,10 +46,6 @@ public class ExPlat internal constructor(
         return activeVariations.getOrPut(experimentIdentifier) {
             getAssignments(NEVER)?.variations?.get(experimentIdentifier) ?: Control
         }
-    }
-
-    override fun refresh(force: Boolean) {
-        refresh(refreshStrategy = if (force) ALWAYS else IF_STALE)
     }
 
     override fun clear() {
--- a/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
+++ b/experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
@@ -83,7 +83,7 @@ public class ExPlat internal constructor(
     }
 
     private suspend fun fetchAssignments() =
-        repository.fetch(platform, experimentIdentifiers, anonymousId).fold(
+        repository.fetch(platform, experimentIdentifiers, anonymousId.orEmpty()).fold(
             onSuccess = {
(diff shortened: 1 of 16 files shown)


experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
```diff
@@ -65,32 +67,29 @@ public class ExPlat internal constructor(
...
 
-    private fun refresh(refreshStrategy: RefreshStrategy) {
-        if (experimentIdentifiers.isNotEmpty()) {
-            getAssignments(refreshStrategy)
-        }
-    }
+    private fun getAssignments(): Assignments? = repository.getCached()
 
-    private fun getAssignments(refreshStrategy: RefreshStrategy): Assignments? {
+    private fun invalidateCache(): Assignments? {
```
iangmaia: Not really clear to me why `invalidateCache()` may return cached assignments?
I mean, even if they're not stale, a function to invalidate them should force (now or on the next fetch of the cache) the cache creation, no? 🤔

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
```diff
@@ -25,37 +22,42 @@ public class ExPlat internal constructor(
...
 
-    init {
-        coroutineScope.launch {
-            refresh()
-        }
-    }
-
-    override fun configure(anonymousId: String?) {
-        clear()
+    override fun initialize(anonymousId: String) {
```
iangmaia, line 25: 💭 

Not for now, and just thinking out loud (and maybe missing more context): perhaps another approach for the API could be to have a builder class (or even a static factory method on the interface?) that will take the `anonymousId` and initialize the classes, so that they wouldn't be able for a client to be create these classes without properly initializing the SDK.

iangmaia: Oh I see that's kinda what `VariationsRepository.create()` does? Can't it also take the `anonymousId`? 🤔

wzieba: > Oh I see that's kinda what VariationsRepository.create() does? Can't it also take the anonymousId? 🤔

Not really, because user can change during the session (logout/sign in). `VariationsRepository` is meant to be a singleton, so `VariationsRepository#create` will be called once for the app lifecycle.

experimentation/src/main/java/com/automattic/android/experimentation/ExPlat.kt
wzieba: Of course you're right, thank you for spotting this 🙌 Addressed in 3dd77fa0b568d43110c3d5ced8af22050659a02b

iangmaia (approved):

PR #246, opened 2024-09-03
Add auth token

wzieba:
### Description

This PR adds a support for providing the oAuthToken when making a request to Experiments API. This token is WPCOM, so it's most likely only useful for Woo/WordPress/Jetpack apps.


iangmaia (approved):
LGTM 👍

PR #265, opened 2025-06-19
Revert "Merge pull request #263 from Automattic/hamorillo/gravatar_user_type"

hamorillo:
After discussing inside the Gravatar team, we realized that we should use the WPCOM `userTypeId` so we don't need to add a new type to the library.  (Sorry for the inconvenience 😖)

This reverts commit c74ea803a5b963c901285e0ac0d2a0c0ffb83e30, reversing changes made to c0193681980308468810a6828882e3742257d3f2.


wzieba (approved):
No worries, I'm glad we aligned early! 🙌

PR #293, opened 2026-01-22
Use Call.Factory instead of OkHttpClient directly

MiSikora:
Using `Call.Factory` instead of `OkHttpClient` allows library consumers to instantiate `OkHttpClient` lazily if desired. This is important because `OkHttpClient` initialization can take upwards of 100 ms on some Android devices, potentially causing ANRs or slowing down app startup.

This change is binary-compatible, as `Call.Factory` is a superinterface of `OkHttpClient`. However, it may be source-incompatible for Kotlin consumers that rely on named parameters, since the constructor parameter was renamed from `okhttpClient` to `callFactory`.

If this is considered a deal-breaker, I can revert the parameter name change. However, I do not think it is a significant issue, as this dependency is typically instantiated only once in most applications, making the migration effort minimal. We can also explicitly call this out in the changelog.

### DIFF
--- a/experimentation/api/experimentation.api
+++ b/experimentation/api/experimentation.api
@@ -29,8 +29,8 @@ public abstract interface class com/automattic/android/experimentation/Variation
 }
 
 public final class com/automattic/android/experimentation/VariationsRepository$Companion {
-	public final fun create (Ljava/lang/String;Ljava/util/Set;Lcom/automattic/android/experimentation/ExperimentLogger;ZLjava/io/File;Lokhttp3/OkHttpClient;Lkotlinx/coroutines/CoroutineScope;Lkotlinx/coroutines/CoroutineDispatcher;)Lcom/automattic/android/experimentation/VariationsRepository;
-	public static synthetic fun create$default (Lcom/automattic/android/experimentation/VariationsRepository$Companion;Ljava/lang/String;Ljava/util/Set;Lcom/automattic/android/experimentation/ExperimentLogger;ZLjava/io/File;Lokhttp3/OkHttpClient;Lkotlinx/coroutines/CoroutineScope;Lkotlinx/coroutines/CoroutineDispatcher;ILjava/lang/Object;)Lcom/automattic/android/experimentation/VariationsRepository;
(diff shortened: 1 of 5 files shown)


wzieba (approved):
Thanks! LGTM

> If this is considered a deal-breaker, I can revert the parameter name change. However, I do not think it is a significant issue, as this dependency is typically instantiated only once in most applications, making the migration effort minimal. We can also explicitly call this out in the changelog.

I think it's totally fine - IMO it's not a deal breaker, especially for internal-only library.

---

Just for a reference, would you mind linking on this PR, a PR from Pocket Casts that will introduce lazy initialization of `OkHttpClient` ? I think it can be a valuable resource for any consumer of the lib.

MiSikora:
> Just for a reference, would you mind linking on this PR, a PR from Pocket Casts that will introduce lazy initialization of `OkHttpClient` ? I think it can be a valuable resource for any consumer of the lib.

Do you mean for me to make a PR with changes in Pocket Casts first against the build from the [7985925](https://github.com/Automattic/Automattic-Tracks-Android/pull/293/commits/79859252e363d46cc60b912c593785c07ebf4cb8) commit and then link it here before merging this PR? If yes, then no problem.

wzieba:
I think that we can do this after integrating the new, stable release to Pocket Casts - not necceseriely against the build from  7985925 commit 👍 

Let me release new version of Tracks now

MiSikora:
So there are two related changes in Pocket Casts:

- [Async Retrofit dispatcher](https://github.com/Automattic/pocket-casts-android/pull/4909). This is needed to ensure that whenever a call is enqueued via a Retrofit instance, the call factory is invoked on a background thread. Otherwise, it is called on the caller thread, which very often is the main thread. However, this only works with functions that are suspending or return `Call<Foo>`. If a Retrofit interface returns something like `Observable`, `CompletableFuture`, etc., you need to make sure that the call adapter you are using dispatches the call on a background thread, or write an adapter for that type.

- [Lazy OkHttp](https://github.com/Automattic/pocket-casts-android/pull/4921). This change ensures that all calls that require creating `OkHttp` or `Cache` are performed on a background thread.

PR #296, opened 2026-03-04
[AI] Add Agent Readiness Configuration

ParaskP7:
Closes: [AINFRA-2083](https://linear.app/a8c/issue/AINFRA-2083)

---

### Description

Add coding agent readiness configuration following the coding agents guidelines (PCYsg-1bzA-p2). This sets up the repository so that coding agents can understand the project structure, build commands, conventions, and gotchas without manual guidance.

#### Changes:
- Add `CLAUDE.md` auto-generated via `/init`, then updated to reference `AGENTS.md` via the `@AGENTS.md` directive
- Add `AGENTS.md` as the single source of truth for all coding agents, documenting project overview, build commands, architecture, testing conventions, and important gotchas
- Add `.aiexclude` to restrict agent access to build artifacts while allowing reports, outputs, and test results
- Update `.gitignore` with local override patterns for popular coding agents (Claude, Codex, Gemini, Cursor, Windsurf) and remove a duplicate IntelliJ block. Junie and Copilot are excluded as Junie's entire directory is local and Copilot has no local override pattern

#### Assessment:

| Category | Status | Reasoning |
|---|---|---|
| **1. Guidelines** | **Done** | `AGENTS.md` documents project overview, build commands, architecture, testing conventions, and gotchas. `CLAUDE.md` points to it via `@AGENTS.md`. `.aiexclude` restricts access to build artifacts and sensitive files. |
| **2. Build & Tests** | **Done** | An agent can build (`./gradlew assembleDebug`), run tests (`./gradlew testDebugUnitTest`), lint (`./gradlew lintDebug`), and check code style (`./gradlew ktlint`) — all documented in `AGENTS.md`. The bootstrap step (`cp gradle.properties-example gradle.properties`) is also documented. CI already runs all of these via Buildkite. |
| **3. E2E Verification** | **TODO** | This is a library, not an app with user-facing UI. There's no E2E framework (no Espresso, UI Automator, Maestro). The `sampletracksapp` exists for manual testing but there's no automated E2E path. The `benchmark` module has instrumented tests but those measure performance, not user-facing behavior. |
| **4. Isolated Env** | **Done** | `.java-version` (21) and Gradle wrapper (9.0.0) are pinned. `gradle.properties-example` serves as the env template. `.gitignore` includes agent local override patterns for all major coding agents. No additional setup scripts or containers needed for this library. |

---

### Testing Steps

- Verify `CLAUDE.md` contains only `@AGENTS.md`
- Verify `AGENTS.md` documents build commands, architecture, testing conventions, and gotchas
- Verify `.aiexclude` excludes build artifacts but allows `reports/`, `outputs/`, and `test-results/`
- Verify `.gitignore` includes local override patterns for agent config files and no longer has duplicate IntelliJ block


.aiexclude
```diff
@@ -0,0 +1,9 @@
+.gradle/
+build/
+!build/reports/
+*/build/
```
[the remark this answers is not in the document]
mokagio: TIL about `.aiexclude`. Neat. Hopefully it takes hold as a wider standard.

Regarding the syntax, Copilot's suggestion is good. From [the docs](https://developer.android.com/studio/gemini/aiexclude#how-write):

> An .aiexclude file follows the same syntax as a [.gitignore file](https://git-scm.com/docs/gitignore).

.gitignore
```diff
@@ -47,3 +41,13 @@ gradle.properties
 # OS X generated file
 .DS_Store
 
+# Agent config local files (Claude, Codex, Gemini, Cursor, Windsurf, etc.)
+# Keep main config files under version control, ignore user-local overrides
+**/AGENTS.local.md
+**/CLAUDE.local.md
```
mokagio, line 47: I didn't know these were a thing. [The docs](https://code.claude.com/docs/en/memory) don't mention supporting a `CLAUDE.local.md` file.

However, it seems to be working

Image

Not sure if it's a compatibility thing that they'll phase out, as it's not documented, or just a standard pattern that will be maintained.

AGENTS.md
```diff
@@ -0,0 +1,70 @@
...
+- **Frameworks:** JUnit 4, AssertJ, Mockito-Kotlin, kotlinx-coroutines-test, MockWebServer
+- **Assertions:** Use AssertJ (`assertThat`), not kotlin.test assertions
+- **Naming:** Backtick-quoted names: `` `given ..., when ..., then ...` ``
+- **Coroutines:** Use `runTest` from `kotlinx.coroutines.test`
+
+## Important Gotchas
+
+- Build files use **Groovy** (`build.gradle`, `settings.gradle`), not Kotlin DSL — except `benchmark/build.gradle.kts` which uses Kotlin DSL
+- Dependency versions are defined as `ext` properties in root `build.gradle`, not a version catalog
+- Lint treats warnings as errors (`warningsAsErrors true`). `AutomatticTracks` uses a lint baseline file
```
mokagio: The first and last seem like stuff the model would notice when reading the code, without needing to be explicit

```suggestion
- Dependency versions are defined as `ext` properties in root `build.gradle`, not a version catalog
```

mokagio (approved):
Looking good. Only nitpicks

.gitignore
ParaskP7: Yea, same @mokagio and thanks for taking a closer look at it, plus testing it! 🥇 

I too just found it myself as I was writing those `Agent Readiness` PRs, of course with AI help and support, and thought, why not adding those in there, wouldn't hurt, would it? Worse case scenario, I thought, we'll stop adding them into more repos (should that not become a standard), and maybe, delete this configuration from those repos that we've added them in already, right? Having said that, usually, this `.gitignore` file becomes a fire and forget file, with ignore files/folders been added in there but rarely removed/edited, only when REALLY needed... 🤷

AGENTS.md
ParaskP7: Good comment @mokagio ! 💯

I was thinking about that too, in the end chose to keep it in there because this `version catalog` is indeed the standard nowadays (years now), and I think AI would try to enforce this standard, maybe creating such a `version catalog` pattern, that is, instead of using the existing legacy `ext` properties pattern this repo uses. What I wanted here, is for AI to avoid doing that, thinking in `version catalog` terms, thus me being explicit giving it instructions on how dependency versions are and should be defined on this repo, being (somehow) explicit that we DON'T WANT `version catalog`. Does that make any sense to you? 💭 (I probably over-complicating this 😊 for sure)

.aiexclude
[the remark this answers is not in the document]
ParaskP7: > Hopefully it takes hold as a wider standard.

🤞 x 🤞 ^ 🤞

.gitignore
mokagio: > why not adding those in there, wouldn't hurt, would it?

Yes, I don't see any problems (other than bloat maybe) in having that in the `.gitignore`. Can always be removed in a future where it's no longer supported.

AGENTS.md
mokagio: Thanks for clarifying. That makes sense to me.

Just to be clear, I am suggesting to remove the "Build files use Groovy" and "Lint treats warnings as errors..." entries because that info is already specified in the document and to keep the version catalog note, which appears in the gotchas for the first time.

mokagio:
_Ship it_

.gitignore
ParaskP7: > (other than bloat maybe)

True @mokagio ! 👍

> Can always be removed in a future where it's no longer supported.

💯

AGENTS.md
ParaskP7: Ah, :facepalm: here @mokagio , and thanks for keeping at it with me... for some reason I read the commit suggestion and that got my brain confused, thinking this comment was about that, while it was about everything else but that, not sure what happened there, apologies... 🫣

> - Build files use **Groovy** (`build.gradle`, `settings.gradle`), not Kotlin DSL — except `benchmark/build.gradle.kts` which uses Kotlin DSL

I would still keep that in there because it is not making sure that AI knows that `Groovy` is the standard in there and not `Kotlin DSL`, except for this one module, just to avoid confusion and help AI chose the right standard. 🤔

> - Lint treats warnings as errors (`warningsAsErrors true`). `AutomatticTracks` uses a lint baseline file

We could definitely remove that, I agree, will do! 👍

mokagio: Sounds good.

ParaskP7: This is now done: b66e127777893aaf34837afe06e6198bf1bd8a73 Cc @mokagio

PR #297, opened 2026-03-31
Fix crash loop caused by corrupted A/B testing cache file

akirk:
## Problem

My Pocket Casts has been stuck in a crash loop. Every launch fails with:

```
FATAL EXCEPTION: DefaultDispatcher-worker-1
java.io.EOFException: End of input
    at com.automattic.android.experimentation.local.CacheDtoJsonAdapter.b
```

After digging through the source, I found that `assignments.json` in `files/experiments/` must have been left empty by an interrupted write. The stack trace makes this clear: `FileNotFoundException` would mean the file doesn't exist; a partial/truncated JSON would give `JsonDataException`. `EOFException: End of input` is specifically what Moshi throws when `fromJson()` receives an empty string — meaning `readText()` succeeded but returned `""`.

The reason I can't recover without a code fix: the file lives in `filesDir`, not `cacheDir`, so clearing the app cache does nothing. Clearing app data would fix the crash but I have starred episodes stored locally that are no longer available in the feed — that data would be gone permanently.

## Fix

Treat an empty or unparseable file as a cache miss, delete it, and let the library re-fetch from the server. I also wrapped the `init` coroutine in `runCatching` so future read failures degrade gracefully rather than crashing the process.

## Follow-up

Worth discussing whether this file should live in `cacheDir` instead of `filesDir` — it's a network cache that can always be re-fetched, and using `cacheDir` would mean Clear Cache recovers users from this situation automatically in the future.

## Test plan

- [x] Added unit test: empty `assignments.json` returns `null` and deletes the file
- [x] Added unit test: corrupted `assignments.json` returns `null` and deletes the file
- [x] Existing tests continue to pass

### DIFF
--- a/experimentation/src/main/java/com/automattic/android/experimentation/local/FileBasedCache.kt
+++ b/experimentation/src/main/java/com/automattic/android/experimentation/local/FileBasedCache.kt
@@ -31,20 +31,33 @@ internal class FileBasedCache(
     init {
         scope.launch {
             withContext(context = dispatcher) {
-                latestMutable = getAssignments()
+                runCatching { latestMutable = getAssignments() }
+                    .onFailure { logger.e("Failed to load cached assignments: $it") }
             }
         }
     }
 
     suspend fun getAssignments(): Assignments? {
         return withContext(dispatcher) {
-            assignmentsFile.takeIf { it.exists() }?.readText()?.let { json: String ->
-                val fromJson = cacheDtoJsonAdapter.fromJson(json) ?: return@let null
-
-                fromJson.assignmentsDto.toAssignments(
-                    fetchedAt = fromJson.fetchedAt,
-                    anonymousId = fromJson.anonymousId,
-                )
+            assignmentsFile.takeIf { it.exists() }?.let { file ->
+                val json = file.readText()
+                if (json.isBlank()) {
+                    logger.e("Cached assignments file is empty, deleting: ${file.path}")
+                    file.delete()
+                    return@withContext null
+                }
+                runCatching { cacheDtoJsonAdapter.fromJson(json) }
+                    .onFailure {
+                        logger.e("Cached assignments file is corrupted, deleting: ${file.path}")
+                        file.delete()
+                    }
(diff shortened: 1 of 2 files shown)


ParaskP7:
👋 @akirk , thanks for reporting and providing us with your fix, we'll take a look as soon as possible! 🙏

> My Pocket Casts has been stuck in a crash loop.

Cc @MiSikora because of the mention of PCAndroid. 😊

ParaskP7:
👋 @akirk !

Thank you so much for your contribution to `Tracks Android` with this PR! 🥇

I have reviewed this and everything LGTM. I also triggered CI on it via a #298 PR, which I just pushed on the main repo, everything works as expected, thanks for making this A/B testing experience better for everyone! 🌟

---

## Things Worth Noting

1. **The `cacheDir` follow-up** you raise is worth a separate issue/PR. Moving `assignments.json` to `cacheDir` would let the OS and "Clear Cache" recover users automatically. But (I think) that's a breaking change (existing cached assignments would be lost on upgrade) so it needs its own discussion.

2. **No `failFast` consideration** — `FileBasedCache` has a `failFast` flag used in `saveAssignments` to throw on directory creation failure. The new `getAssignments` error handling ignores `failFast` — it always degrades gracefully. I think this is actually the right behavior (reading a corrupt cache should never crash regardless of `failFast`), but it's worth being aware of the asymmetry.

---

Overall, this is a 👍 from my side and if @MiSikora doesn't have anything more to share, let's merge this and create a new version of `Tracks Android`. 🚀

ParaskP7 (approved):

MiSikora (approved):
Looks good, no objections.

ParaskP7:
👋 @akirk @MiSikora and FYI that version [6.0.8](https://github.com/Automattic/Automattic-Tracks-Android/releases/tag/6.0.8) is ready for you to use.

akirk:
Just to circle back, this must have made it into PocketCasts Android 8.9 because now it loads again without crash :) Thanks!

